PR #2, opened 2018-06-06
Start implementing input

mlgthatsme:


Source/AliveLib/Input.cpp
```diff
@@ -0,0 +1,39 @@
...
+	if (!keyState)
+		return false;
+	sInputKeyStates_BD2F60[a1] = keyState & 0x80;
+	return true;
+}
+ALIVE_FUNC_IMPLEX(0x4EDD40, Abe_IsVKPressed_4EDD40, true);
+
+
+// Unimplemented
+ALIVE_FUNC_IMPLEX(0x460280, InputGetJoystickInput_460280, false);
```
paulsapps: you can use `ALIVE_FUNC_NOT_IMPL(0x4949F0, void CC(), Main_4949F0);` for unimpl stuff, sometimes its cleaner if its not partly impl'd

Source/AliveLib/Input.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+ALIVE_FUNC_IMPLEX(0x4EDD60, Input_InitKeyStateArray_4EDD60, true);
+
+
+// Returns true if a key was just pressed down.
+bool __cdecl Abe_IsVKPressed_4EDD40(int a1)
+{
+	char keyState = sInputKeyStates_BD2F60[a1];
+	if (!keyState)
+		return false;
```
paulsapps: ideally no single line if's :)

Source/AliveLib/Input.cpp
```diff
@@ -0,0 +1,39 @@
...
+// Zeros the input key state array.
+void CC Input_InitKeyStateArray_4EDD60()
+{
+	memset(sInputKeyStates_BD2F60, 0, 256u);
+}
+ALIVE_FUNC_IMPLEX(0x4EDD60, Input_InitKeyStateArray_4EDD60, true);
+
+
+// Returns true if a key was just pressed down.
+bool __cdecl Abe_IsVKPressed_4EDD40(int a1)
```
paulsapps: a1 should probably be called "key"

Source/AliveLib/Input.cpp
```diff
@@ -0,0 +1,39 @@
...
+ALIVE_VAR(1, 0x5C2F00, int, sJoystickID_5C2F00, 0);
+
+// -- Functions -- //
+
+// Zeros the input key state array.
+void CC Input_InitKeyStateArray_4EDD60()
+{
+	memset(sInputKeyStates_BD2F60, 0, 256u);
+}
+ALIVE_FUNC_IMPLEX(0x4EDD60, Input_InitKeyStateArray_4EDD60, true);
```
paulsapps: For `true` I normally have a file level #define INPUT_IMPL true so that you can change it to false when trying to track down regressions

Source/AliveLib/Input.hpp
```diff
@@ -0,0 +1 @@
+void __cdecl InputGetJoystickInput_460280(float *X1, float *Y1, float *X2, float *Y2, DWORD *Buttons);
```
paulsapps: __cdecl has a define of CC for calling convention so it can easily be compiled out if need be

mlgthatsme:
Try that!

PR #4, opened 2018-10-10
impl Slurg::Clear_Slurg_Step_Watch_Points_449A90

JayPeet:


Source/AliveLib/Slurg.cpp
```diff
@@ -0,0 +1,29 @@
+#include "stdafx.h"
```
paulsapps: This cpp and hpp needs adding to CMakeLists.txt otherwise it won't get compiled/linked

Source/AliveLib/Slurg.cpp
```diff
@@ -0,0 +1,29 @@
...
+char Slurg::sSlurg_Step_Watch_Points_Count_5BD4DC[2];
+Slurg_Step_Watch_Points Slurg::sSlurg_Step_Watch_Points_5C1B28[2];
+
+void CC Slurg::Clear_Slurg_Step_Watch_Points_449A90()
+{
+	sSlurg_Step_Watch_Points_Idx_5C1C08 = sSlurg_Step_Watch_Points_Idx_5C1C08 == 0;
+	sSlurg_Step_Watch_Points_Count_5BD4DC[sSlurg_Step_Watch_Points_Idx_5C1C08] = 0;
+
+	Slurg_Step_Watch_Points* point = &sSlurg_Step_Watch_Points_5C1B28[sSlurg_Step_Watch_Points_Idx_5C1C08];
+	point->field_0_points[0].field_0_xPos = 0;
```
paulsapps: you can use a range based for and C++ default in here, e.g

for (auto& point : sSlurg_Step_Watch_Points_5C1B28[sSlurg_Step_Watch_Points_Idx_5C1C08].field_0_points)
{
 point = {};
}

Source/AliveLib/Slurg.cpp
```diff
@@ -0,0 +1,29 @@
+#include "stdafx.h"
+#include "Slurg.hpp"
+
+WORD Slurg::sSlurg_Step_Watch_Points_Idx_5C1C08;
+char Slurg::sSlurg_Step_Watch_Points_Count_5BD4DC[2];
```
paulsapps: Need to use ALIVE_VAR/ALIVE_ARY on these - might need to make them globals in the CPP for that to work

paulsapps:
Added a couple of comments - also all tabs -> spaces ;)

PR #78, opened 2019-06-05
Slig.hpp and Slig.cpp numbering error FIXES

AvengingGibbons:
Does this mean i get true programmer role?


mouzedrift:
> Does this mean i get true programmer role?

no

Issue #121, opened 2019-08-07
Evil Fart has wrong color


UltraStars3000:
**Found in build:** Never worked 

**Description of issue:**
When you fart, the evil fart hasn't the same color as OG.

**Steps to repro:**
Load the save and fart.
[EvilFart.zip](https://github.com/AliveTeam/alive_reversing/files/3479071/EvilFart.zip)

**Function:** EvilFart::vUpdate_423100

PR #124, opened 2019-08-09
Fart colour fix.

Varrok:
🎷 🐛


paulsapps:
This is for #121

Varrok:
This is also for #125

paulsapps:
And for #125

PR #146, opened 2019-08-23
impl Scrab::AI_ChasingEnemy_1_4A6470

UltraStars3000:
Probably a lot of things are done badly or totally wrong. Would need a code review sorry ^^'

### DIFF
--- a/Source/AliveLib/Scrab.cpp
+++ b/Source/AliveLib/Scrab.cpp
@@ -685,477 +685,432 @@ __int16 Scrab::AI_Patrol_0_4AA630()
 
 __int16 Scrab::AI_ChasingEnemy_1_4A6470()
 {
-    __int16 result;
-
-    Scrab *pScrabToFight = Scrab::FindScrabToFight_4A4E20();
+    Scrab *pScrabToFight = FindScrabToFight_4A4E20();
     if (pScrabToFight)
     {
         SetBrain(&Scrab::AI_Fighting_2_4A5840);
         field_124_fight_target_obj_id = pScrabToFight->field_8_object_id;
         field_108_next_motion = eScrabMotions::M_Stand_0_4A8220;
-        result = AI_Fighting::eState2_LookingForOpponent_0;
+        return AI_Fighting::eState2_LookingForOpponent_0;
     }
-    else
+    auto pObj = static_cast<BaseAliveGameObject*>(sObjectIds_5C1B70.Find_449CF0(field_120_obj_id));
+    if (!pObj || field_6_flags.Get(BaseGameObject::eDead) || static_cast<int>(sGnFrame_5C1B84) > field_14C && !CanSeeAbe_4A51A0(pObj))
     {
-        auto pObj = static_cast<BaseAliveGameObject*>(sObjectIds_5C1B70.Find_449CF0(field_120_obj_id));
-        if (!pObj || field_6_flags.Get(BaseGameObject::eDead) || static_cast<int>(sGnFrame_5C1B84) > field_14C && !Scrab::CanSeeAbe_4A51A0(pObj))
+        field_120_obj_id = -1;
+        field_108_next_motion = eScrabMotions::M_Stand_0_4A8220;
+        ToPatrol_4AA600();
+        return AI_Patrol::eState0_ToMoving_0;
+    }
+    if (pObj->field_114_flags.Get(Flags_114::e114_Bit8_bInvisible))
+    {
+        field_120_obj_id = -1;
+        field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
+        ToPatrol_4AA600();
+        return AI_Patrol::eState0_UNKNOWN_9;
+    }
+    if (CanSeeAbe_4A51A0(pObj))
+    {
+        field_14C = sGnFrame_5C1B84 + field_148_attack_duration;
+    }
+    if (Event_Is_Event_In_Range_422C30(kEventAbeOhm, field_B8_xpos, field_BC_ypos, -1) && field_11C_sub_state != 26)
+    {
+        field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
+        return AI_ChasingEnemy::eState1_Panic_4;
+    }
+    LiftPoint* pLiftPoint = static_cast<LiftPoint*>(pObj);
+    if (!pObj || pObj->field_4_typeId != Types::eLiftPoint_78 || pLiftPoint->vOnAnyFloor_461920() || field_11C_sub_state == 4)
+    {
+        if (gMap_5C3030.GetDirection_4811A0(
+            field_C2_lvl_number,
+            field_C0_path_number,
+            field_B8_xpos,
+            field_BC_ypos) >= CameraPos::eCamCurrent_0)
         {
-            field_120_obj_id = -1;
-            field_108_next_motion = eScrabMotions::M_Stand_0_4A8220;
-            ToPatrol_4AA600();
-            result = AI_Patrol::eState0_ToMoving_0;
+            MusicController::sub_47FD60(8, this, 0, 0);
         }
-        else if (pObj->field_114_flags.Get(Flags_114::e114_Bit8_bInvisible))
+        switch (field_11C_sub_state)
         {
-            field_120_obj_id = -1;
-            field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
-            ToPatrol_4AA600();
-            result = AI_Patrol::eState0_UNKNOWN_9;
-        }
-        else
-        {
-            if (Scrab::CanSeeAbe_4A51A0(pObj))
+        case AI_ChasingEnemy::eState1_UNKNOWN_0:
+            field_150 = sGnFrame_5C1B84 + field_128_attack_delay;
+            return AI_ChasingEnemy::eState1_Idle_1;
+        case AI_ChasingEnemy::eState1_Idle_1:
+            if (!field_106_current_motion)
             {
-                field_14C = sGnFrame_5C1B84 + field_148_attack_duration;
+                field_194 = LastSpeak_4A56F0();
+                if (field_1A2 < field_1A0
+                    && LastSpeak_4A56F0() == GameSpeakEvents::eUnknown_53
+                    || LastSpeak_4A56F0() == GameSpeakEvents::eUnknown_54)
+                {
+                    return AI_ChasingEnemy::eState1_UNKNOWN_15;
+                }
             }
-            if (Event_Is_Event_In_Range_422C30(kEventAbeOhm, field_B8_xpos, field_BC_ypos, -1) && field_11C_sub_state != 26)
+            if (!vIsFacingMe_4254A0(pObj))
             {
-                field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
-                result = AI_ChasingEnemy::eState1_Panic_4;
+                if (gMap_5C3030.Is_Point_In_Current_Camera_4810D0(field_C2_lvl_number, field_C0_path_number, field_B8_xpos, field_BC_ypos, 0))
+                {
+                    field_108_next_motion = eScrabMotions::M_Turn_3_4A91A0;
+                }
+                else
+                {
+                    field_106_current_motion = eScrabMotions::M_Turn_3_4A91A0;
+                    field_108_next_motion = -1;
+                    MapFollowMe_408D10(TRUE);
+                }
+                return AI_ChasingEnemy::eState1_Turning_3;
+            }
+            if (vIsObjNearby_4253B0(ScaleToGridSize_4498B0(field_CC_sprite_scale) / FP_FromInteger(5), pObj)
+                && vOnSameYLevel_425520(pObj))
+            {
+                if (!vIsObjNearby_4253B0(ScaleToGridSize_4498B0(field_CC_sprite_scale), pObj))
+                {
+                    field_108_next_motion = eScrabMotions::M_AttackLunge_37_4AA0B0;
+                    return AI_ChasingEnemy::eState1_Attacking_8;
+                }
+                field_108_next_motion = eScrabMotions::M_LegKick_38_4AA120;
+                return AI_ChasingEnemy::eState1_Attacking_8;
+            }
+            else if (CanSeeAbe_4A51A0(pObj) && static_cast<int>(sGnFrame_5C1B84) >= field_150)
+            {
+                if (Handle_SlamDoor_or_EnemyStopper_4A4830(field_C4_velx, 0))
+                {
+                    return field_11C_sub_state;
+                }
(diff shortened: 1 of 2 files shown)


Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+    {
+        SetBrain(&Scrab::AI_Fighting_2_4A5840);
+        field_124_fight_target_obj_id = pScrabToFight->field_8_object_id;
+        field_108_next_motion = eScrabMotions::M_Stand_0_4A8220;
+        result = 0;
+    }
+    else
+    {
+        BaseGameObject *pObj = sObjectIds_5C1B70.Find_449CF0(field_120_obj_id);
+        BaseAliveGameObject *v6 = (BaseAliveGameObject *)pObj;
```
paulsapps: get rid of v6 and use `auto pObj = static_cast<BaseAliveGameObject*>(sObjectIds_5C1B70.Find_449CF0(field_120_obj_id));`

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+        if (!pObj || field_6_flags.Get(BaseGameObject::eDead) || (signed int)sGnFrame_5C1B84 > field_14C && !Scrab::CanSeeAbe_4A51A0((BaseAliveGameObject *)pObj))
```
paulsapps: use static_cast<int>(gnframe)

paulsapps: remove BaseAliveGameObject *) cast

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                field_14C = sGnFrame_5C1B84 + field_148_attack_duration;
+            }
+            if (Event_Is_Event_In_Range_422C30(kEventAbeOhm, field_B8_xpos, field_BC_ypos, -1) && field_11C_sub_state != 26)
+            {
+                field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
+                result = 4;
+            }
+            else
+            {
+                pObj = sObjectIds_5C1B70.Find_449CF0(field_110_id);
```
paulsapps: rather than assigning to pObj just static cast directly to the assignment of pLift

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                            {
+                                field_108_next_motion = eScrabMotions::M_Turn_3_4A91A0;
+                            }
+                            else
+                            {
+                                field_106_current_motion = eScrabMotions::M_Turn_3_4A91A0;
+                                field_108_next_motion = -1;
+                                BaseAliveGameObject::MapFollowMe_408D10(1);
+                            }
+                            result = 3;
```
paulsapps: result can be removed from this function and just return the value directly with `return 3`;

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                        field_BC_ypos) >= CameraPos::eCamCurrent_0)
+                    {
+                        MusicController::sub_47FD60(8, this, 0, 0);
+                    }
+                    switch (field_11C_sub_state)
+                    {
+                    case 0:
+                        result = 1;
+                        field_150 = sGnFrame_5C1B84 + field_128_attack_delay;
+                        return result;
```
paulsapps: Same as below, the var result is redundant, just `return 1;` here

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                    case 1:
+                        if (!field_106_current_motion)
+                        {
+                            field_194 = static_cast<short>(Scrab::LastSpeak_4A56F0());
```
paulsapps: Scrab:: is redundant, the cast is also redundant you can use the GameSpeakEvents type (or whatever it returns)

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                            if (field_1A2 < field_1A0 
+                                && Scrab::LastSpeak_4A56F0() == GameSpeakEvents::eUnknown_53
```
paulsapps: ditto for redundant Scrab::

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                        {
+                            if (gMap_5C3030.Is_Point_In_Current_Camera_4810D0(field_C2_lvl_number, field_C0_path_number, field_B8_xpos, field_BC_ypos, 0))
+                            {
+                                field_108_next_motion = eScrabMotions::M_Turn_3_4A91A0;
+                            }
+                            else
+                            {
+                                field_106_current_motion = eScrabMotions::M_Turn_3_4A91A0;
+                                field_108_next_motion = -1;
+                                BaseAliveGameObject::MapFollowMe_408D10(1);
```
paulsapps: can use TRUE instead of 1 here

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                            if (!vIsObjNearby_4253B0(ScaleToGridSize_4498B0(field_CC_sprite_scale), v6))
+                            {
+                                field_108_next_motion = eScrabMotions::M_AttackLunge_37_4AA0B0;
+                                result = 8;
+                                return result;
+                            }
+                            field_108_next_motion = eScrabMotions::M_LegKick_38_4AA120;
+                            result = 8;
+                        }
+                        else if (Scrab::CanSeeAbe_4A51A0(v6) && (signed int)sGnFrame_5C1B84 >= field_150)
```
paulsapps: static_cast

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                            {
+                                result = field_11C_sub_state;
+                                return result;
+                            }
+                            result = 2;
+                            field_108_next_motion = eScrabMotions::M_Run_2_4A89C0;
+                        }
+                        else
+                        {
+                            if (Math_NextRandom() >= 0x10u || (signed int)(sGnFrame_5C1B84 - field_154) <= 60)
```
paulsapps: use decimal instead of hex 0x10 here, and static_cast

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                            {
+                                if (!vIsObjNearby_4253B0(ScaleToGridSize_4498B0(field_CC_sprite_scale) / FP_FromInteger(7), v6)
+                                    || !v6->field_114_flags.Get(Flags_114::e114_Bit8_bInvisible) 
+                                    || v6->field_4_typeId != Types::eScrab_112)
+                                {
+                                    if (vIsObjNearby_4253B0(ScaleToGridSize_4498B0(field_CC_sprite_scale) * FP_FromInteger(3), v6)
+                                        && field_106_current_motion == eScrabMotions::M_Run_2_4A89C0
+                                        && vOnSameYLevel_425520(v6))
+                                    {
+                                        if (BaseAliveGameObject::WallHit_408750(field_CC_sprite_scale * FP_FromInteger(45), v6->field_B8_xpos - field_B8_xpos))
```
paulsapps: BaseAliveGameObject:: is redundant ClassName:: is always redundant if the function isn't `static`

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +628,483 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                        }
+                        if (field_194 != 53)
+                        {
+                            result = field_11C_sub_state;
+                            return result;
+                        }
+                        field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
+                        result = 13;
+                        break;
+                    case 16:
```
paulsapps: you can probably add an enum for these case values since you've figured out some names for them?

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,477 @@ __int16 Scrab::AI_Patrol_0_4AA630()
 
 __int16 Scrab::AI_ChasingEnemy_1_4A6470()
 {
-    NOT_IMPLEMENTED();
-    return 0;
+    __int16 result;
+
+    Scrab *pScrabToFight = Scrab::FindScrabToFight_4A4E20();
```
paulsapps: Scrab:: redundant here

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,477 @@ __int16 Scrab::AI_Patrol_0_4AA630()
 
 __int16 Scrab::AI_ChasingEnemy_1_4A6470()
 {
-    NOT_IMPLEMENTED();
-    return 0;
+    __int16 result;
```
paulsapps: This var can be factored out

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,477 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
-    return 0;
+    __int16 result;
+
+    Scrab *pScrabToFight = Scrab::FindScrabToFight_4A4E20();
+    if (pScrabToFight)
+    {
+        SetBrain(&Scrab::AI_Fighting_2_4A5840);
+        field_124_fight_target_obj_id = pScrabToFight->field_8_object_id;
+        field_108_next_motion = eScrabMotions::M_Stand_0_4A8220;
+        result = AI_Fighting::eState2_LookingForOpponent_0;
```
paulsapps: if you return result here you can remove the else block

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,477 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+    }
+    else
+    {
+        auto pObj = static_cast<BaseAliveGameObject*>(sObjectIds_5C1B70.Find_449CF0(field_120_obj_id));
+        if (!pObj || field_6_flags.Get(BaseGameObject::eDead) || static_cast<int>(sGnFrame_5C1B84) > field_14C && !Scrab::CanSeeAbe_4A51A0(pObj))
+        {
+            field_120_obj_id = -1;
+            field_108_next_motion = eScrabMotions::M_Stand_0_4A8220;
+            ToPatrol_4AA600();
+            result = AI_Patrol::eState0_ToMoving_0;
```
paulsapps: if you return result here you can remove the else block

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,477 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+        }
+        else if (pObj->field_114_flags.Get(Flags_114::e114_Bit8_bInvisible))
+        {
+            field_120_obj_id = -1;
+            field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
+            ToPatrol_4AA600();
+            result = AI_Patrol::eState0_UNKNOWN_9;
```
paulsapps: if you return result here you can remove the else block

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,477 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+        else
+        {
+            if (Scrab::CanSeeAbe_4A51A0(pObj))
+            {
+                field_14C = sGnFrame_5C1B84 + field_148_attack_duration;
+            }
+            if (Event_Is_Event_In_Range_422C30(kEventAbeOhm, field_B8_xpos, field_BC_ypos, -1) && field_11C_sub_state != 26)
+            {
+                field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
+                result = AI_ChasingEnemy::eState1_Panic_4;
```
paulsapps: if you return result here you can remove the else block

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,477 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+            }
+            else
+            {
+                LiftPoint* pLiftPoint = nullptr;
```
paulsapps: can assign in 1 line instead of 2

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,432 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+    {
+        field_14C = sGnFrame_5C1B84 + field_148_attack_duration;
+    }
+    if (Event_Is_Event_In_Range_422C30(kEventAbeOhm, field_B8_xpos, field_BC_ypos, -1) && field_11C_sub_state != 26)
+    {
+        field_108_next_motion = eScrabMotions::M_HowlBegin_26_4A9DA0;
+        return AI_ChasingEnemy::eState1_Panic_4;
+    }
+    LiftPoint* pLiftPoint = static_cast<LiftPoint*>(pObj);
+    if (!pObj || pObj->field_4_typeId != Types::eLiftPoint_78 || pLiftPoint->vOnAnyFloor_461920() || field_11C_sub_state == 4)
```
paulsapps: I think you should use pLiftPoint instead of pObj in the if statement as the are actually the same thing

paulsapps: I think if you invert this whole if statement you can remove the else block first with an early return

Source/AliveLib/Scrab.cpp
```diff
@@ -627,8 +685,432 @@ __int16 Scrab::AI_Patrol_0_4AA630()
...
+                if (Math_NextRandom() >= 16u || static_cast<int>(sGnFrame_5C1B84 - field_154) <= 60)
+                {
+                    return AI_ChasingEnemy::eState1_Idle_1;
+                }
+                field_108_next_motion = eScrabMotions::M_Shriek_30_4A9EA0;
+                field_154 = sGnFrame_5C1B84;
+                return AI_ChasingEnemy::eState1_Shriek_14;
+            }
+        case AI_ChasingEnemy::eState1_Running_2:
+            field_194 = LastSpeak_4A56F0();
```
paulsapps: I think all of case 2 should go into its own helper function because its huge and will become easier to refactor some parts of it

PR #167, opened 2019-09-09
Renamed variables in Bone and BoneBag Classes

THEONLYDarkShadow:
Renamed variables in Bone and BoneBag classes to describe what they do. Also cleaned up logic around field_120 that was not needed.

### DIFF
--- a/Source/AliveLib/Bone.cpp
+++ b/Source/AliveLib/Bone.cpp
@@ -36,14 +36,14 @@ Bone* Bone::ctor_4112C0(FP xpos, FP ypos, __int16 countId)
     field_C4_velx = FP_FromInteger(0);
     field_C8_vely = FP_FromInteger(0);
     field_6_flags.Clear(BaseGameObject::eInteractive);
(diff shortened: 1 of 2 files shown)


Source/AliveLib/Bone.cpp
```diff
@@ -736,24 +736,16 @@ void BoneBag::vUpdate_412880()
...
-                SFX_Play_46FBA0(29u, 24, Math_RandomRange_496AB0(-2400, -2200));
-            }
-        }
-    }
-    else
-    {
-        field_120 = 1;
+      if (Math_NextRandom() < 40 || field_122_force_play_sound)
+      {
+        field_122_force_play_sound = 0;
```
paulsapps: This seems to be changing behaviour as 

  field_120 = 0;	   

Is no longer set to 0?

THEONLYDarkShadow: All references to `field_120` have been removed since the variable is redundant and not needed. From what I've tested, there seems to be no change in behaviour.

PR #178, opened 2019-09-21
add bug-template-v2.md

mouzedrift:


paulsapps:
shouldn't it also have original game bug label in there too? :gibbons:

paulsapps:
I thought this was a template but OG bugs but its the reverse :oddtism:

PR #194, opened 2019-09-28
Fleech Sfx function reversed.

Varrok:
🐛 👅

### DIFF
--- a/Source/AliveLib/Fleech.cpp
+++ b/Source/AliveLib/Fleech.cpp
@@ -16,6 +16,7 @@
 #include "ObjectIds.hpp"
 #include "PlatformBase.hpp"
 #include "PsxDisplay.hpp"
+#include "Sfx.hpp"
 
 ALIVE_VAR(1, 0x5BC20C, BYTE, sFleechRandomIdx_5BC20C, 0);
 ALIVE_VAR(1, 0x5BC20E, short, sFleechCount_5BC20E, 0);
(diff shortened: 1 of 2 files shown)


Source/AliveLib/Fleech.cpp
```diff
@@ -1846,10 +1872,75 @@ void Fleech::ToIdle_42E520()
...
+const SfxDefinition getSfxDef(FleechSound effectId)
 {
-    NOT_IMPLEMENTED();
-    return 0;
+    return stru_5518E0[static_cast<int>(effectId)];
+}
+
+int Fleech::Sound_430520(FleechSound soundId)
+{
+    SfxDefinition effectDef = getSfxDef(soundId);
```
paulsapps: const

Source/AliveLib/Fleech.cpp
```diff
@@ -1846,10 +1872,75 @@ void Fleech::ToIdle_42E520()
...
+    __int16 defaultSndIdxVol = effectDef.field_3_default_volume;
```
paulsapps: const

Source/AliveLib/Fleech.cpp
```diff
@@ -1846,10 +1872,75 @@ void Fleech::ToIdle_42E520()
...
+    if (soundId == FleechSound::Unknown_14)
+    {
+        int rndIdx = Math_RandomRange_496AB0(14, 16);
```
paulsapps: const

PR #195, opened 2019-09-29
Math_496F70()

Varrok:
No guarantees that the current version is 1:1 with OG 😔

### DIFF
--- a/Source/AliveLib/Math.cpp
+++ b/Source/AliveLib/Math.cpp
@@ -223,29 +223,32 @@ EXPORT FP CC Math_496F70(FP value1, FP value2)
         switchScenario += 2;
     }
     switchScenario += value2abs < value1abs;
+    FP aux = {};
     switch (switchScenario)
     {
     case 1:
-        const FP aux = value2abs / value1abs;
+        aux = value2abs / value1abs;
         return (FP_FromInteger(2) - aux) * FP_FromInteger(32);
     case 2:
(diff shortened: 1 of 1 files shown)


Varrok:
When I said no guarantees, I meant things like removed case 0

Source/AliveLib/Math.cpp
```diff
@@ -207,91 +207,51 @@ int CC Math_Distance_496EB0(int x1, int y1, int x2, int y2)
...
-        v3 = &value2;
+        value1abs = -value1;
+        switchScenario += 4;
     }
-    else
+    if (value2 < FP_FromInteger(0))
     {
-        v16 = -value2;
-        v3 = &v16;
+        value2abs = -value2;
```
paulsapps: there are actual abs() / FP_Abs() functions for this

Source/AliveLib/Math.cpp
```diff
@@ -207,91 +207,51 @@ int CC Math_Distance_496EB0(int x1, int y1, int x2, int y2)
...
+        switchScenario += 2;
     }
-
-    const FP v4 = *v3;
-    FP* v5 = nullptr;
-    if (value1 >= FP_FromInteger(0))
+    switchScenario += value2abs < value1abs;
```
paulsapps: this should be made explicit instead of boolean implicit casting, use value2abs < value1abs ? 1 : 0

PR #234, opened 2019-10-18
Fleech fix

d3nwah:
🐛 🔫


Source/AliveLib/Fleech.cpp
```diff
@@ -1616,11 +1616,12 @@ void Fleech::RenderEx_42C5A0(int** ot)
                 pScreenManager_5BB5F4->field_3A_idx
             );
         }
+        auto bufferIdx = gPsxDisplay_5C1130.field_C_buffer_index;
```
paulsapps: I don't think you should use this local as its kind of redundant, not const and not used else where in the func

Issue #239, opened 2019-10-21
the status screen in the pause menu uses the word "has" instead of "have"


mouzedrift:
**Found in build:** 1849

**Steps to repro:** 
1. start the game
2. open the pause menu
3. select "status"

**Function:** ``PauseMenu::Update_48FD80``

# reversed
![Exoddus_NoStubs_w60kDj0Wyp](https://user-images.githubusercontent.com/43358824/67234969-720f3c80-f446-11e9-930a-2baf83e53027.png)

# OG
![Exoddus_NoStubs_wFoMib1soQ](https://user-images.githubusercontent.com/43358824/67234976-75a2c380-f446-11e9-97b2-8bdcd11242cf.png)

PR #240, opened 2019-10-21
Pause menu has/have bug fix.

d3nwah:


paulsapps:
You should tag this with  #239 so we know what its fixing

d3nwah:
> You should tag this with #239 so we know what its fixing

I hereby tag this with #239

PR #245, opened 2019-10-22
DDCheat::Menu_Movies_416000()

d3nwah:


Source/AliveLib/DDCheat.cpp
```diff
@@ -131,7 +135,46 @@ void DDCheat::Menu_Teleport_415E20()
...
+    }
+
+    if (Path_Get_FMV_Record_460F70(gMap_5C3030.sCurrentLevelId_5C3030, sDDCheat_MovieSelectIdx_5BBFF0)->field_4_id <= 0 )
+    {
+        sDDCheat_MovieSelectIdx_5BBFF0 = 1;
+    }
+
+    if (field_38_input_pressed & InputCommands::eDown)
+    {
+        auto movieId = Path_Get_FMV_Record_460F70(gMap_5C3030.sCurrentLevelId_5C3030, sDDCheat_MovieSelectIdx_5BBFF0)->field_4_id;
```
paulsapps: this local is pointless, might as well just have the -- on the end of id

Source/AliveLib/DDCheat.cpp
```diff
@@ -131,7 +135,46 @@ void DDCheat::Menu_Teleport_415E20()
...
+        movieId--;
+    }
+    if (field_38_input_pressed & InputCommands::eUp)
+    {
+        FmvInfo *movieToPlayInfo = Path_Get_FMV_Record_460F70(gMap_5C3030.sCurrentLevelId_5C3030, sDDCheat_MovieSelectIdx_5BBFF0);
+        DWORD pos;
```
paulsapps: = 0

Source/AliveLib/DDCheat.cpp
```diff
@@ -131,7 +135,46 @@ void DDCheat::Menu_Teleport_415E20()
...
+                pos,
+                movieToPlayInfo->field_6_flags & 1,
+                movieToPlayInfo->field_8,
+                movieToPlayInfo->field_A_volume
+            );
+        }
+    }
+    FmvInfo* fmvInfo = &sPathData_559660.paths[static_cast<int>(gMap_5C3030.sCurrentLevelId_5C3030)].field_4_pFmvArray[sDDCheat_MovieSelectIdx_5BBFF0];
+    DDCheat::DebugStr_4F5560("\n<- Movie -> %d %d %s \n", sDDCheat_MovieSelectIdx_5BBFF0, fmvInfo->field_4_id, fmvInfo->field_0_pName);
+    field_20 += 6;
```
paulsapps: can this field be renamed? what is it? += 6 seems suspect?

Source/AliveLib/DDCheat.cpp
```diff
@@ -131,7 +135,45 @@ void DDCheat::Menu_Teleport_415E20()
...
+    if (field_38_input_pressed & InputCommands::eLeft)
+    {
+        sDDCheat_MovieSelectIdx_5BBFF0--;
+    }
+    else if (field_38_input_pressed & InputCommands::eRight)
+    {
+        sDDCheat_MovieSelectIdx_5BBFF0++;
+    }
+
+    if (Path_Get_FMV_Record_460F70(gMap_5C3030.sCurrentLevelId_5C3030, sDDCheat_MovieSelectIdx_5BBFF0)->field_4_id <= 0 )
```
paulsapps: you know it might bet better to put Path_Get_FMV_Record_460F70(gMap_5C3030.sCurrentLevelId_5C3030, sDDCheat_MovieSelectIdx_5BBFF0) into a local because its looked up 3 times in here

d3nwah: I disagree, for two reasons:
-the first call might return a different value than the other two, as the if block can modify sDDCheat_MovieSelectIdx_5BBFF0 
-the latter two calls are independent, and won't happen simultaneously (unless someone holds both up and down arrow, which I'd consider OG bug)

PR #246, opened 2019-10-23
Abe_SFX_457EC0() cleanup

d3nwah:


Source/AliveLib/Abe.hpp
```diff
@@ -176,6 +176,51 @@ enum class AbeSay : __int16
     eSympathy_28 = 28,
 };
 
+const enum class sAbeSFXList_555250_Tag : unsigned __int8
```
paulsapps: const makes no sense here and why _Tag?

Source/AliveLib/Abe.cpp
```diff
@@ -1456,11 +1456,11 @@ void Abe::Update_449DC0()
                 if (field_128.field_18_say == AbeSay::eSympathy_28)
                 {
                     // This one has another volume for whatever reason
-                    Abe_SFX_457EC0(static_cast<unsigned char>(field_128.field_18_say), 80, 0, this);
+                    Abe_SFX_457EC0(static_cast<sAbeSFXList_555250_Tag>(field_128.field_18_say), 80, 0, this);
```
paulsapps: I think that field_18_say should be typed to sAbeSFXList_555250_Tag

Source/AliveLib/Abe.cpp
```diff
@@ -9956,100 +9956,115 @@ void Abe::ExitShrykull_45A9D0(__int16 bResetRingTimer)
     }
 }
 
-// TODO: Clean up
-EXPORT void CC Abe_SFX_457EC0(unsigned __int8 idx, __int16 volume, int pitch, BaseAliveGameObject* pHero)
+void playAbeSFX(sAbeSFXList_555250_Tag idx, __int16 volume, int pitch)
```
paulsapps: mark static

Source/AliveLib/Abe.hpp
d3nwah: Because those are labels for sAbeSFXList_555250[] sounds

Issue #155, opened 2019-09-01
some sligs try to beat mudokons at an empty spot


mouzedrift:
**Found in build:** 1633

**Steps to repro:** load the save and sneak to the right screen (you might need more than 1 try)

**Save:** [test.zip](https://github.com/AliveTeam/alive_reversing/files/3563033/test.zip)

**Cause:** Slig::FindBeatTarget_4BD070

mouzedrift:
![Exoddus_NoStubs_xfWN4xK7FD](https://user-images.githubusercontent.com/43358824/65378640-af53a380-dcbb-11e9-8bf4-b40db8de46a1.png)
so he beats the mudokon at the first and the second position but he shouldnt do it at the second position. in the OG game he beats the mudokon at the first position and goes to the next shadow. (he stops in the OG game at the second position for a really short time and continues walking but he doesnt beat him) maybe OG bug?

mouzedrift:
fixed 🚀

mouzedrift:
ugh this is still happening at different positions feelsbadman 

**save:** [slig2.zip](https://github.com/AliveTeam/alive_reversing/files/3693570/slig2.zip)

**video:** https://streamable.com/dktxm

PR #259, opened 2019-10-28
Slig beat fix.

d3nwah:
#155 fix

### DIFF
--- a/Source/AliveLib/Slig.cpp
+++ b/Source/AliveLib/Slig.cpp
@@ -2356,6 +2356,39 @@ void Slig::M_LiftDown_50_4B3960()
     field_106_current_motion = MoveLift_4B3990(FP_FromInteger(4));
 }
 
+
+inline PSX_RECT MakeRectFromFP(FP x, FP y, FP w, FP h)
+{
+    PSX_RECT r = {};
+    r.x = FP_GetExponent(x);
+    r.w = FP_GetExponent(w);
+    r.y = FP_GetExponent(y);
+    r.h = FP_GetExponent(h);
+    return r;
+}
+
+inline PSX_RECT MakeMinMaxRect(FP x, FP y, FP w, FP h, bool flipToMaxMin = false)
+{
+    if (flipToMaxMin)
+    {
+        return MakeRectFromFP(
+            std::max(x, w),
+            std::max(y, h),
+            std::min(x, w),
+            std::min(y, h)
+        );
+    }
+    else
+    {
+        return MakeRectFromFP(
+            std::min(x, w),
+            std::min(y, h),
+            std::max(x, w),
+            std::max(y, h)
+        );
+    }
+}
+
 void Slig::M_Beat_51_4B6C00()
 {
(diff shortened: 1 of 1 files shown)


Source/AliveLib/Slig.cpp
```diff
@@ -6592,11 +6581,14 @@ BaseAliveGameObject* Slig::FindBeatTarget_4BD070(int /*a2*/, int gridBlocks)
             PSX_RECT bRect = {};
             pObj->vGetBoundingRect_424FD0(&bRect, 1);
 
-            if (pObj->field_10C_health > FP_FromInteger(0) &&
+            if (hitRect.w <= bRect.w &&
```
paulsapps: why can this use PSX_Rects_overlap_no_adjustment?

Source/AliveLib/Slig.cpp
```diff
@@ -6592,11 +6581,14 @@ BaseAliveGameObject* Slig::FindBeatTarget_4BD070(int /*a2*/, int gridBlocks)
...
+                hitRect.x >= bRect.x &&
+                hitRect.y >= bRect.y &&
+                hitRect.h <= bRect.h &&
                 pObj->field_D6_scale == field_D6_scale &&
                 !IsInInvisibleZone_425710(pObj) &&
                 !pObj->field_114_flags.Get(Flags_114::e114_Bit8_bInvisible) &&
-                PSX_Rects_overlap_no_adjustment(&hitRect, &bRect))
+                pObj->field_10C_health > FP_FromInteger(0))
```
paulsapps: didn't this make more sense to be first? Do the cheap calc before the expensive ones?

Source/AliveLib/Slig.cpp
```diff
@@ -6551,33 +6551,22 @@ void Slig::PlayerControlRunningSlideStopOrTurn1_4B85D0()
...
 
-    const FP rectX = field_20_animation.field_4_flags.Get(AnimFlags::eBit5_FlipX) ? field_B8_xpos - kGridBlocksScaled : field_B8_xpos + kGridBlocksScaled;
-
-    const PSX_RECT hitRect = MakeRectFromFP(
-        rectX,
-        field_BC_ypos,
-        rectX,
-        field_BC_ypos - k2Scaled
-    );
+    const FP hitRectX = field_20_animation.field_4_flags.Get(AnimFlags::eBit5_FlipX) ? field_B8_xpos - kGridBlocksScaled : field_B8_xpos + kGridBlocksScaled;
```
paulsapps: I think this should still have some FP_Rect helper to init a rect since its what this is doing ?

Source/AliveLib/Slig.cpp
d3nwah: PSX_Rects_overlap_no_adjustment will return false if any of the params are exactly the same in both Rects, unlikely as it is.

Source/AliveLib/Slig.cpp
d3nwah: It does, however we're not at the refactoring stage yet. All I'm doing here is make it IDA compliant

PR #321, opened 2019-12-27
Input stuff

d3nwah:


Source/AliveLib/Input.cpp
```diff
@@ -910,82 +932,54 @@ void NewParseSettingsIni()
             {
                 printf("Value: %s is %s\n", param[0].c_str(), param[1].c_str());
```
paulsapps: replace with LOG_INFO

Source/AliveLib/Input.cpp
```diff
@@ -910,82 +932,54 @@ void NewParseSettingsIni()
...
-                                        }
-                                    }
-                                    else // int
-                                    {
-                                        *reinterpret_cast<int*>(s.data) = atoi(param[1].c_str());
-                                    }
-
-                                    foundConfig = true;
-                                    break;
+                                    *reinterpret_cast<bool*>(s.data) = true;
```
paulsapps: Existing but change the structure to just have an int and then add an AsInt() AsBool() method or something, or make it a union, anything but this :)

Source/AliveLib/Input.cpp
```diff
@@ -908,84 +939,56 @@ void NewParseSettingsIni()
...
                     {
-                        bool foundConfig = false;
-                        // custom config loader code
+                        if (param[0] == s.name)
                         {
-                            for (const IniCustomSaveEntry& s : gCustomSaveEntries)
+                            if (s.isBool)
                             {
-                                if (param[0] == s.name)
+                                if (param[1] == "true")
```
paulsapps: why not just  *s.data.boolVal = param[1] == "true" ?

PR #324, opened 2020-01-18
Minecar State1 refactor

d3nwah:
Refactor of the State_1() of Minecar's update function.

There's a bug while holding the movement key and going round the corners, the minecar doesn't turn in every angle scenario

### DIFF
--- a/Source/AliveLib/MineCar.cpp
+++ b/Source/AliveLib/MineCar.cpp
@@ -979,52 +979,14 @@ void MineCar::State_1()
     field_C4_velx = FP_FromInteger(0);
     field_C8_vely = FP_FromInteger(0);
 
-    // TODO: Refactor
-    bool donyEnterRightBlock = false;
-    if (!sInputObject_5BD4E0.isPressed(sInputKey_Right_5550D0))
-    {
-        if (!sInputObject_5BD4E0.isPressed(field_1D4_previous_input))
-        {
-            donyEnterRightBlock = true;
-        }
-
-        if ((unsigned __int16) field_1D6_continue_move_input != sInputKey_Right_5550D0)
-        {
-            donyEnterRightBlock = true;
-        }
-
-        if (field_1BC == 3)
-        {
-            donyEnterRightBlock = true;
-        }
-
-        if (field_1BC == 0)
-        {
-            donyEnterRightBlock = true;
-        }
-
-        if ((field_1D4_previous_input) == (unsigned __int16) sInputKey_Left_5550D4)
-        {
-            donyEnterRightBlock = true;
-        }
-    }
-
-    if (donyEnterRightBlock ||
-        WallHit_408750(k60Scaled * FP_FromDouble(0.5), k12Adjusted + FP_FromInteger(1)) ||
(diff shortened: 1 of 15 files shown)


Source/AliveLib/MineCar.cpp
```diff
@@ -294,7 +294,7 @@ int CC MineCar::CreateFromSaveState_467740(const BYTE* pBuffer)
     pMineCar->field_118_tlvInfo = pState->field_4C_tlvInfo;
 
     pMineCar->field_11C_state = pState->field_50_state;
-    pMineCar->field_1BC = pState->field_52;
+    pMineCar->field_turnDirection_1BC = static_cast<MineCarDirs>(pState->field_52);
```
paulsapps: can't the type of field_52 be  also changed to MineCarDirs and be renamed to represent what it actually is?

Source/AliveLib/MineCar.cpp
```diff
@@ -933,14 +938,19 @@ void MineCar::State_0()
     }
 }
 
+FP mineCarHeightUnscaled = FP_FromInteger(60);
```
paulsapps: these should be const, and probably static too

Source/AliveLib/MineCar.cpp
```diff
@@ -972,206 +982,184 @@ void MineCar::State_1()
...
+    if ((sInputObject_5BD4E0.isPressed(inputKey) ||
+        (sInputObject_5BD4E0.isPressed(field_1D4_previous_input) &&
+        (unsigned __int16) field_1D6_continue_move_input == inputKey &&
+        field_1D4_previous_input != (unsigned __int16) inputKey &&
+        field_turnDirection_1BC != MineCarDirs::eUp_3 &&
+        field_turnDirection_1BC != MineCarDirs::eDown_0))
+        && !(WallHit_408750(mineCarHeight * FP_FromDouble(0.5), mineCarWidthAdjusted + FP_FromInteger(1)) &&
+        WallHit_408750(mineCarHeight - FP_FromInteger(8), mineCarWidthAdjusted + FP_FromInteger(1))))
+    {
+        FP rayCast1 = (kGridSize / FP_FromInteger(4)) + field_B8_xpos;
```
paulsapps: nearly all these locals can be const, also would it not be clearer to just pass them directly maybe?

Source/AliveLib/MineCar.cpp
```diff
@@ -972,206 +982,184 @@ void MineCar::State_1()
...
-        {
-            if (WallHit_408750(FP_FromInteger(1), kGridSize + k12Scaled + FP_FromInteger(4))||
-                WallHit_408750(k60Scaled + FP_FromInteger(1), kGridSize + k12Scaled + FP_FromInteger(4)))
-            {
-                if (hitX > field_B8_xpos)
-                {
-                    Move_46E640(20836u, FP_FromInteger(0), -k5Scaled, (unsigned short)sInputKey_Up_5550D8, 2, 0);
-                }
-            }
+        const FP funArg1 = FP_FromInteger(1);
```
paulsapps: these names kinda suck, also const/or directly pass as function arg?

Issue #363, opened 2020-04-20
Evilfart doesn't explode while zapped


d3nwah:
Found in build: master  97f0a337204b0f18974771d9ef6ec16ce11ddb97
Last working build number: presumably never

Description of issue:
Evilfart doesn't explode while zapped. It just gets purple, making me feel blue.

Steps to repro:
Fly with a fart through a zapfield
![image](https://user-images.githubusercontent.com/6751254/79800295-3d7c2f00-835c-11ea-824f-e7a1055ac49e.png)

d3nwah:
[handy_savegame.zip](https://github.com/AliveTeam/alive_reversing/files/4506202/handy_savegame.zip)

mouzedrift:
**Function:** ``EvilFart::vUpdate_423100``

PR #374, opened 2020-04-30
#363 fix.

d3nwah:


d3nwah:
Also some renames, minor cleanups

Source/AliveLib/EvilFart.hpp
```diff
@@ -5,6 +5,13 @@
 
 enum class LevelIds : __int16;
 
+enum class FartStates : __int16
+{
+    e0_IDLE = 0,
```
paulsapps, line 10: All the other enums are eName_Number

Source/AliveLib/FlyingSlig.cpp
```diff
@@ -908,7 +908,7 @@ void FlyingSlig::Movement_4396E0()
     field_B8_xpos += field_C4_velx;
     field_BC_ypos += field_C8_vely;
 
-    sub_408C40();
+    setActiveCameraDelayedFromDir_408C40();
```
paulsapps, line 911: all the other methods are MethodName not methodName

Source/AliveLib/EvilFart.cpp
```diff
@@ -316,39 +316,38 @@ __int16 EvilFart::VTakeDamage_423B70(BaseGameObject* pFrom)
     return 1;
 }
 
+void EvilFart::vUpdate_REAL_423100()
```
paulsapps, line 319: can be removed as it was only for testing?

PR #454, opened 2020-07-09
fix #378

### LINKED ISSUE #378: Greeters don't turn when shot
**Description of issue:**
When shot by a slig, a greeter should face it instead of not turning a all, which is the case here.

**Steps to repro:**
Load the save: [Greeter.zip](https://github.com/AliveTeam/alive_reversing/files/4618869/Greeter.zip)
Start shooting with the slig.

**Function:** `Greeter::vTakeDamage_447C20`

UltraStars3000:
And made ElectricWall to its own case

### DIFF
--- a/Source/AliveLibAE/Greeter.cpp
+++ b/Source/AliveLibAE/Greeter.cpp
@@ -511,6 +511,9 @@ EXPORT signed __int16 Greeter::vTakeDamage_447C20(BaseGameObject* pFrom)
         field_20_animation.field_4_flags.Clear(AnimFlags::eBit3_Render);
         BlowUp_447E50();
         return 1;
+
+    default:
+        return 1;
     }
 }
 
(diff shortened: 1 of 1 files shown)


Source/AliveLibAE/Greeter.cpp
```diff
@@ -472,8 +513,7 @@ EXPORT signed __int16 Greeter::vTakeDamage_447C20(BaseGameObject* pFrom)
         return 1;
 
     default:
-        // Seems that this can be eElectricWall_39 - maybe others ??
-        LOG_WARNING("This might be wrong - greeter default damage case, can this only be a slig ?? type is " << static_cast<int>(pFrom->field_4_typeId));
+        LOG_WARNING("This might be wrong - greeter default damage case, type is " << static_cast<int>(pFrom->field_4_typeId));
         if (static_cast<BaseAnimatedWithPhysicsGameObject*>(pFrom)->field_20_animation.field_10_frame_delay <= 0)
```
paulsapps, line 517: this part can be removed now since you moved the code to an explicit case statement

Issue #546, opened 2020-08-09
AO Rev Bug - Abe can't walk when hes in the background area


mouzedrift:
Steps to repro:
- start the game and follow the main path until you're in the background area

Function: Abe::sub_422FC0

PR #556, opened 2020-08-25
#546 fix.

d3nwah:
🥵

### DIFF
--- a/Source/AliveLibAO/Abe.cpp
+++ b/Source/AliveLibAO/Abe.cpp
@@ -1614,8 +1614,8 @@ void Abe::sub_422FC0()
     }
 
     // TODO: Check mask is correct
-    const int mask = field_BC_sprite_scale != FP_FromDouble(0.5) ? 0x11 : 0x10;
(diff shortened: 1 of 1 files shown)


Source/AliveLibAO/Abe.cpp
```diff
@@ -1614,8 +1614,8 @@ void Abe::sub_422FC0()
     }
 
     // TODO: Check mask is correct
-    const int mask = field_BC_sprite_scale != FP_FromInteger(0) ? 1 : 0x10;
-    if (field_F4_pLine && (mask & (1 << field_F4_pLine->field_8_type)))
+    const int mask = field_BC_sprite_scale != FP_FromDouble(0.5) ? 0x11 : 0x10;
```
paulsapps, line 1617: I think this should be 0x01 and 0x10 ya know, it would match what AE does as well...

d3nwah: It should, but it isn't what IDA says...

AE doesn't have the (1 << field_F4_pLine->field_8_type) part too

PR #558, opened 2020-09-01
fix OG bug where abe doesn't transform into shrykull when you use gamespeak and then immediately chant

mouzedrift:
https://www.youtube.com/watch?v=hHioSCNnAps

### DIFF
--- a/Source/AliveLibAE/Abe.cpp
+++ b/Source/AliveLibAE/Abe.cpp
@@ -9125,6 +9125,7 @@ short Abe::DoGameSpeak_45AB70(int input)
     short nextState = -1;
     if ((sInputObject_5BD4E0.field_0_pads[sCurrentControllerIndex_5C1BBE].field_0_pressed & eChant) == eChant)
     {
+        // Fixes an OG bug where abe doesn't transform into shrykull when you immediately chant after using gamespeak
         if (!(field_168_ring_pulse_timer && field_16C_bHaveShrykull))
         {
             field_124_gnFrame = sGnFrame_5C1B84 + 90;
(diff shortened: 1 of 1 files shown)


d3nwah:
So when Abe speaks and the player immediately tries to chant, what happens? Does he transform or is the chant input skipped? @mouzedrift

mouzedrift:
he transforms as expected

d3nwah (approved):

d3nwah:
Before merging it, we should reverse the appropriate AO function (DoGameSpeak_42F5C0) and use the exact fix there as well.

paulsapps:
I don't think this bug happens in AO does it?

mouzedrift:
nope it doesn't

Source/AliveLibAE/Abe.cpp
```diff
@@ -9125,10 +9125,13 @@ short Abe::DoGameSpeak_45AB70(int input)
     short nextState = -1;
     if ((sInputObject_5BD4E0.field_0_pads[sCurrentControllerIndex_5C1BBE].field_0_pressed & eChant) == eChant)
     {
-        field_124_gnFrame = sGnFrame_5C1B84 + 90;
-        SND_SEQ_PlaySeq_4CA960(SeqId::MudokonChant1_10, 0, 1);
-        field_120_state = 0;
-        nextState = eAbeStates::State_112_Chant_45B1C0;
+        if (!(field_168_ring_pulse_timer && field_16C_bHaveShrykull))
```
paulsapps, line 9128: I think it might be worth adding a comment on this new if statement saying something like:

// Fix OG bug where abe doesn't transform into shrykull when you use gamespeak and then immediately chant 

Since this will be the first game play changing bug we've fixed that is OG?

mouzedrift: can do but i think we already had some OG fixes/logic changes (excluding crash bug fixes)

PR #572, opened 2020-09-06
Paramite brains + Brain_Death_448BF0

UltraStars3000:
There might be some things to check, I've tried my best to sort out if/else statements

### DIFF
--- a/Source/AliveLibAO/Paramite.cpp
+++ b/Source/AliveLibAO/Paramite.cpp
@@ -22,6 +22,7 @@
 #include "CameraSwapper.hpp"
 #include "PlatformBase.hpp"
 #include "LiftPoint.hpp"
+#include <Particle.hpp>
 
 void Paramite_ForceLink() {}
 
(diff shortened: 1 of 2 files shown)


Source/AliveLibAO/Paramite.cpp
```diff
@@ -984,31 +985,50 @@ __int16 Paramite::Brain_44DD70()
...
+        field_C2_g -= 2;
+        field_C4_b -= 2;
+        if (static_cast<int>(gnFrameCount_507670) < field_114 - 24 && !(static_cast<int>(gnFrameCount_507670) % 5))
+        {
+            New_Particles_419A80(
+                (FP_FromInteger(Math_RandomRange_450F20(-24, 24)) * field_BC_sprite_scale) + field_A8_xpos,
+                field_AC_ypos - FP_FromInteger(6),
+                field_BC_sprite_scale / FP_FromInteger(2),
+                2, 0);
+            SFX_Play_43AE60(0x60u, 25, FP_GetExponent(FP_FromInteger(2200) * field_BC_sprite_scale), 0);
```
paulsapps, line 1019: probably better to use decimal instead of 0x60u here ?

PR #574, opened 2020-09-06
Impl paramite state(s)

mouzedrift:


Source/AliveLibAO/Paramite.cpp
```diff
@@ -1059,7 +1052,7 @@ void Paramite::State_0_Idle_44B900()
         }
         else
         {
-            MusicController::sub_443810(MusicController::MusicTypes::eType12, this, 0, 0);
+            MusicController::sub_443810(MusicController::MusicTypes::eType11, this, 0, 0);
```
paulsapps, line 1055: this used to be 12?

Source/AliveLibAO/Paramite.cpp
```diff
@@ -1211,9 +1204,223 @@ void Paramite::State_2_Walking_44B9E0()
     }
 }
 
+int sRunningTable_4BBCC0[14] =
```
paulsapps, line 1207: should be const int, the others should be as well if they're not

Source/AliveLibAO/Paramite.cpp
```diff
@@ -1211,9 +1204,223 @@ void Paramite::State_2_Walking_44B9E0()
...
+            else
+            {
+                MusicController::sub_443810(MusicController::MusicTypes::eType11, this, 0, 0);
+            }
+        }
+    }
+
+    field_142 = 0;
+
+    if (gMap_507BA8.GetDirection(
```
paulsapps, line 1409: this GetDirection / ChaseBrain and music playing seems like an inlined function its repeated a lot

Source/AliveLibAO/Paramite.cpp
mouzedrift: this was previously 12 but should've been 11

Source/AliveLibAO/Paramite.cpp
mouzedrift: forgot about that one the others are fine

Source/AliveLibAO/Paramite.cpp
mouzedrift: created a new helper func for it

Source/AliveLibAO/Paramite.cpp
```diff
@@ -1211,9 +1223,167 @@ void Paramite::State_2_Walking_44B9E0()
...
+    if (field_10_anim.field_92_current_frame == 3)
+    {
+        SFX_Play_43AE60(0x21u, 45, -600, 0);
+        if (field_FE_next_state == eParamiteStates::State_2_Walking_44B9E0)
+        {
+            field_FC_current_motion = eParamiteStates::State_8_WalkRunTransition_44C790;
+            field_FE_next_state = -1;
+        }
+
+        if (!field_FE_next_state
```
paulsapps, line 1282: !field_FE_next_state is checking that state is 0

Source/AliveLibAO/Paramite.cpp
```diff
@@ -1211,9 +1223,167 @@ void Paramite::State_2_Walking_44B9E0()
...
+            VOnTrapDoorOpen();
+
+            field_E8_LastLineYPos = field_AC_ypos;
+            field_FC_current_motion = eParamiteStates::State_6_Hop_44CB20;
+            field_FE_next_state = -1;
+            field_B8_vely = -FP_FromDouble(7.2) * field_BC_sprite_scale;
+            field_AC_ypos += field_B8_vely;
+            field_F4_pLine = nullptr;
+
+            if (field_142)
```
paulsapps, line 1317: check this, looks like bug?

Source/AliveLibAO/Paramite.cpp
```diff
@@ -1211,9 +1223,167 @@ void Paramite::State_2_Walking_44B9E0()
...
+            ToHop_44B660();
+        }
+
+        if (!field_142)
+        {
+            MapFollowMe_401D30(1);
+            field_142 = 1;
+        }
+
+        SetMusic();
```
paulsapps, line 1381: check this, was there a missing return here? otherwise its gonna overwrite field_142 again and call set music again

Source/AliveLibAO/Paramite.cpp
```diff
@@ -1211,9 +1223,167 @@ void Paramite::State_2_Walking_44B9E0()
...
+    MoveOnLine_44B740();
+
+    if (field_FC_current_motion != eParamiteStates::State_3_Running_44C070)
+    {
+        return;
+    }
+
+    if (field_10_anim.field_92_current_frame == 3)
+    {
+        SFX_Play_43AE60(0x21u, 45, -600, 0);
```
paulsapps, line 1275: might wanna switch 0x21u to decimal so replacing with an enum later is easier

PR #580, opened 2020-09-12
State_30_HopMid_4264D0()

d3nwah:
hop

### DIFF
--- a/Source/AliveLibAO/Abe.cpp
+++ b/Source/AliveLibAO/Abe.cpp
@@ -2967,9 +2967,103 @@ void Abe::State_29_HopBegin_4267B0()
     NOT_IMPLEMENTED();
 }
 
-void Abe::State_30_HopMid_4264D0()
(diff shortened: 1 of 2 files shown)


Source/AliveLibAO/Abe.cpp
```diff
@@ -2967,11 +2967,103 @@ void Abe::State_29_HopBegin_4267B0()
...
+                    case 4:
+                    case 32:
+                    case 36:
+                    {
+                        Abe_SFX_2_42A220(6u, 0, 0x7FFF, this);
+                        field_F4_pLine = pLine;
+                        field_B8_vely = FP_FromInteger(0);
+                        field_FC_current_motion = eAbeStates::State_31_HopLand_426940;
+
+                        PSX_RECT pRect;
```
paulsapps, line 3022: should be called rect not pRect as it isn't a pointer, also needs = {}

PR #594, opened 2020-09-15
impl some abe funcs

mouzedrift:


Source/AliveLibAO/Abe.cpp
```diff
@@ -2039,10 +2040,107 @@ void Abe::FallOnBombs_4231B0()
     }
 }
 
+__int16 Abe::CrouchOrPushWall(const FP gridSize)
+{
+    if (!WallHit_401930(field_BC_sprite_scale * FP_FromInteger(50), gridSize))
```
paulsapps, line 2045: seems like it would make more sense as:

if (WallHit(50))
{
  if (WallHit(20))
  {
    crouch
    return 0;
   }
}
return 1;
?

Source/AliveLibAO/Abe.cpp
```diff
@@ -2039,10 +2040,107 @@ void Abe::FallOnBombs_4231B0()
...
+            field_B4_velx = gridSize / FP_FromInteger(9);
+            field_FC_current_motion = eAbeStates::State_6_WalkBegin_424300;
+        }
+
+        if (field_B4_velx > FP_FromInteger(0))
+        {
+            return CrouchOrPushWall(gridSize);
+        }
+
+        gridSize = FP_FromInteger(0);
```
paulsapps, line 2100: this can't be a gridSize if its zero

Source/AliveLibAO/Abe.hpp
```diff
@@ -282,6 +285,8 @@ class Abe : public BaseAliveGameObject
 
     EXPORT void sub_422FC0();
 
+    EXPORT eAbeStates MoveLiftUpOrDown_42F190(FP ySpeed);
```
paulsapps, line 288: probably want to return __int16 instead for now

Source/AliveLibAO/Abe.cpp
```diff
@@ -2039,10 +2040,107 @@ void Abe::FallOnBombs_4231B0()
...
+        field_FC_current_motion = eAbeStates::State_21_StandToCrouch_427F40;
+        return 1;
+    }
+}
+
 __int16 Abe::ToLeftRightMovement_422AA0()
 {
-    NOT_IMPLEMENTED();
-    return 0;
+    field_B8_vely = FP_FromInteger(0);
```
paulsapps, line 2066: maybe check AE to see if the logic can be re-arranged a bit? Seems like flipX and some other stuff can be grouped a bit better for readablity? If not I can re-arrange it later

PR #595, opened 2020-09-16
abe and elum funcs

mouzedrift:


Source/AliveLibAO/Elum.cpp
```diff
@@ -1014,7 +1099,45 @@ void Elum::State_3_WalkLoop_412C90()
...
+            field_10_anim.Set_Animation_Data_402A40(field_10_anim.field_18_frame_table_offset, 0);
+            if (sControlledCharacter_50767C == this)
+            {
+                sActiveHero_507678->SyncToElum_42D850(field_FC_current_motion);
+            }
+
+        }
+        else
+        {
+            field_118 = 0;
```
paulsapps, line 1132: Looks like a ToIdle inlined?

Source/AliveLibAO/Elum.cpp
```diff
@@ -1191,17 +1357,83 @@ void Elum::State_13_RunTurnToWalk_4147C0()
...
+    else
+    {
+        field_170_flags |= 0x20u;
+    }
+
+    if (field_10_anim.field_4_flags.Get(AnimFlags::eBit18_IsLastFrame))
+    {
+        if (!ToNextState_4120F0())
+        {
+            field_118 = 0;
```
paulsapps, line 1374: An Elum ToIdle inlined func perhaps?

Source/AliveLibAO/Elum.cpp
```diff
@@ -1191,17 +1357,83 @@ void Elum::State_13_RunTurnToWalk_4147C0()
...
+    }
+    else
+    {
+        field_170_flags |= 0x20u;
+    }
+    if (field_10_anim.field_4_flags.Get(AnimFlags::eBit18_IsLastFrame))
+    {
+        if (!ToNextState_4120F0())
+        {
+            field_118 = 0;
```
paulsapps, line 1400: ToIdle ?

Source/AliveLibAO/Elum.cpp
```diff
@@ -1191,17 +1357,83 @@ void Elum::State_13_RunTurnToWalk_4147C0()
...
+    {
+        field_170_flags |= 0x20u;
+    }
+
+    if (field_10_anim.field_4_flags.Get(AnimFlags::eBit18_IsLastFrame))
+    {
+        field_FE_next_state = eElumStates::State_10_Yell_4158E0;
+        if (!ToNextState_4120F0())
+        {
+            field_118 = 0;
```
paulsapps, line 1428: ToIdle?

Source/AliveLibAO/Elum.cpp
```diff
@@ -1454,17 +1686,125 @@ void Elum::State_29_BeesStruggling_412A90()
...
+                field_AC_ypos = hitY;
+                break;
+            default:
+                break;
+            }
+        }
+
+        if (field_AC_ypos - field_E8_LastLineYPos > FP_FromInteger(2))
+        {
+            field_118 = 0x22666;
```
paulsapps, line 1763: Probably a FP type - can be changed later though

Source/AliveLibAO/Elum.cpp
```diff
@@ -313,6 +313,20 @@ __int16 Elum::VTakeDamage_411020(BaseGameObject* pFrom)
...
+void Elum::toKnockback()
+{
+    field_B4_velx = FP_FromInteger(0);
+    field_FC_current_motion = eElumStates::State_50_Knockback_415DC0;
+
+    if (field_F4_pLine)
+    {
+        MoveOnLine_412580(0);
+    }
+    field_A8_xpos -= field_B4_velx;
```
paulsapps, line 325: This seems a bit odd - I guess that MoveOnLine_412580 changes field_B4_velx otherwise it always subtracts zero which does nothing?

Source/AliveLibAO/Abe.cpp
```diff
@@ -5361,7 +5365,27 @@ void Abe::State_136_ElumMountEnd_42E110()
...
+    {
+        while (!ResourceManager::FreeResource_455550(field_1A4_resources.res[58]))
+        {
+            // Empty
+        }
+
+        field_1A4_resources.res[58] = {};
+        field_104_pending_resource_count = 3;
+
+        ResourceManager::LoadResourceFile_4551E0("ABEBSIC.BAN", reinterpret_cast<ResourceManager::TLoaderFn>(OnResourceLoaded_4019A0), reinterpret_cast<Camera*>(this), 0);
```
paulsapps, line 5378: Use the template overload:

 ResourceManager::LoadResourceFile("ABENOELM.BND", BaseAliveGameObject::OnResourceLoaded_4019A0, this);

Then you can drop the casts, you can remove the last 0 argument as it defaults to in this template helper, same for the other LoadResourceFile calls in here

Source/AliveLibAO/Abe.cpp
```diff
@@ -5361,7 +5365,27 @@ void Abe::State_136_ElumMountEnd_42E110()
...
+        ResourceManager::LoadResourceFile_4551E0("ANEPRMNT.BAN", reinterpret_cast<ResourceManager::TLoaderFn>(OnResourceLoaded_4019A0), reinterpret_cast<Camera*>(this), 0);
+        ResourceManager::LoadResourceFile_4551E0("ABENOELM.BND", reinterpret_cast<ResourceManager::TLoaderFn>(OnResourceLoaded_4019A0), reinterpret_cast<Camera*>(this), 0);
+
+        if (!ResourceManager::GetLoadedResource_4554F0(ResourceManager::Resource_Animation, 48, FALSE, FALSE))
+        {
+            ++field_104_pending_resource_count;
```
paulsapps, line 5384: I usually post increment these: field_104_pending_resource_count++;

Source/AliveLibAO/Abe.cpp
```diff
@@ -5371,7 +5395,20 @@ void Abe::State_138_ElumUnmountEnd_42E390()
...
 {
-    NOT_IMPLEMENTED();
+    if (!field_104_pending_resource_count && field_10_anim.field_4_flags.Get(AnimFlags::eBit18_IsLastFrame))
+    {
+        while (!ResourceManager::FreeResource_455550(field_1A4_resources.res[61]))
+        {
+            // Empty
+        }
+
+        field_1A4_resources.res[61] = {};
```
paulsapps, line 5405: might be better to use = nullptr; if its only clearing 1 element rather than all of them

Source/AliveLibAO/Abe.cpp
```diff
@@ -5371,7 +5395,20 @@ void Abe::State_138_ElumUnmountEnd_42E390()
...
+        field_104_pending_resource_count = 1;
+
+        ResourceManager::LoadResourceFile_4551E0("ABEWELM.BND", reinterpret_cast<ResourceManager::TLoaderFn>(OnResourceLoaded_4019A0), reinterpret_cast<Camera*>(this), 0);
```
paulsapps, line 5408: can use the template helper - same comment as the other above

Source/AliveLibAO/Elum.cpp
mouzedrift: you're right it subtracts zero which does nothing

Source/AliveLibAO/Elum.cpp
```diff
@@ -1191,17 +1349,65 @@ void Elum::State_13_RunTurnToWalk_4147C0()
...
+    if (field_10_anim.field_4_flags.Get(AnimFlags::eBit18_IsLastFrame))
+    {
+        if (!ToNextState_4120F0())
+        {
+            ToIdle();
+        }
+    }
 }
 
 void Elum::State_16_Speak_414980()
```
paulsapps, line 1391: this function is exactly the same as State_15_Speak_4148F0, also State_14_Speak_414860 is almost identical. You could add a helper method void Speak(int soundId); that removes the duplicated code as the first soundId passed to Abe_SFX_2_42A220() is the only change between these 3?

PR #605, opened 2020-09-19
Moved SlurgSpawner class to its own set of files. Named fields and structures in the SlogHut, Slurg and SlurgSpawner classes.

THEONLYDarkShadow:
Moved `SlurgSpawner` class to its own set of files. Named fields and structures in the `SlogHut`, `Slurg` and `SlurgSpawner` classes.

### DIFF
--- a/Source/AliveLibAE/SlurgSpawner.cpp
+++ b/Source/AliveLibAE/SlurgSpawner.cpp
@@ -1,0 +1,85 @@
+#include "stdafx.h"
+#include "SlurgSpawner.hpp"
+#include "DDCheat.hpp"
+#include "Function.hpp"
+#include "stdlib.hpp"
+#include "SwitchStates.hpp"
+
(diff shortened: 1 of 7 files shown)


Source/AliveLibAE/SlurgSpawner.cpp
```diff
@@ -0,0 +1,85 @@
...
+    BaseGameObject_ctor_4DBFA0(TRUE, 0);
+    SetVTable(this, 0x547704);
+
+    field_4_typeId = Types::eSlurgSpawner_128;
+    field_20_tlvInfo = tlvInfo;
+    field_24_slurg_tlv = pTlv;
+    field_28_path_data = pTlv->field_18_path_data;
+    field_3E_delay_counter = 0;
+    field_40_spawned_count = pTlv->field_1_unknown;
+    if ((field_40_spawned_count & 0x80u) != 0)
```
paulsapps, line 19: I think that is actually supposed to be field_40_spawned_count  < 0

PR #645, opened 2020-09-27
impl some funcs

mouzedrift:


Source/AliveLibAO/BaseBomb.cpp
```diff
@@ -6,14 +6,141 @@
...
 START_NS_AO
 
 ALIVE_VAR(1, 0x4FFA4C, short, word_4FFA4C, 0);
 
 void BaseBomb::VUpdate_417580()
 {
-    NOT_IMPLEMENTED();
+    PSX_RECT rect = {};
+
+    Event_Broadcast_417220(2, this);
```
paulsapps, line 21: should be an enum value for this

Source/AliveLibAO/BaseBomb.cpp
```diff
@@ -6,14 +6,141 @@
...
+    Event_Broadcast_417220(14, this);
```
paulsapps, line 22: and this one

PR #650, opened 2020-09-28
Abe::BulletDamage_4220B0.

d3nwah:


Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,228 @@ __int16 Abe::TryMountElum_42E600()
     return eAbeStates::State_0_Idle_423520;
 }
 
+void Abe::BulletDamage_4220B0(Bullet* pBullet)
+{
+    PSX_RECT rect = {};
+    VGetBoundingRect_418120(&rect, 1);
```
paulsapps, line 2266: drop the address else isn't gonna be a virtual call

Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,228 @@ __int16 Abe::TryMountElum_42E600()
...
+
+    enum class ShootKind
+    {
+        eEverythingElse_0 = 0,
+        eHanging_1 = 1,
+        eRolling_2 = 2,
+    };
+
+    if (field_FC_current_motion != eAbeStates::State_137_ElumUnmountBegin_42E2B0
```
paulsapps, line 2275: wouldn't it be better to invert this + return early so the rest of the func isn't so indented?

Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,228 @@ __int16 Abe::TryMountElum_42E600()
...
+                    pBlood->ctor_4072B0(
+                        field_A8_xpos,
+                        pBullet->field_1C_ypos,
+                        yPos,
+                        FP_FromInteger(0),
+                        field_BC_sprite_scale,
+                        50);
+                }
+
+                if (shootKind != ShootKind::eEverythingElse_0)
```
paulsapps, line 2331: this would be easier to understand inverted

paulsapps: actually can't this be a switch ?

Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,228 @@ __int16 Abe::TryMountElum_42E600()
...
+                        ElumKnockForward_42E780(1);
+                    }
+                    if (field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX) != pBullet->field_20 > FP_FromInteger(0))
+                    {
+                        field_FC_current_motion = eAbeStates::State_128_KnockForward_429330;
+                    }
+                    field_108_bMotionChanged = 1;
+                    field_106_shot = 0;
+
+                    field_B4_velx = field_BC_sprite_scale * FP_FromDouble(7.8);
```
paulsapps, line 2369: can't this go in the else? seems strange to calc it and then overwrite it sometimes

Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,228 @@ __int16 Abe::TryMountElum_42E600()
...
+            case BulletType::Type_2:
+            {
+                if (field_BC_sprite_scale == FP_FromDouble(0.5))
+                {
+                    field_106_shot = 0;
+                    field_100_health = FP_FromInteger(1);
+                    return;
+                }
+
+                Path_TLV *pPath = nullptr;
```
paulsapps, line 2386: you should call this pTlv as its called in all other usages

Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,228 @@ __int16 Abe::TryMountElum_42E600()
...
+                        pPath,
+                        FP_FromInteger(rect.x),
+                        FP_FromInteger(rect.y),
+                        FP_FromInteger(rect.x),
+                        FP_FromInteger(rect.y)
+                    );
+                    if (!pPath)
+                        break;
+
+                    if (pPath->field_4_type != 83)
```
paulsapps, line 2399: use the enum type, looks like ZCover and I think this is the inlined check for ZCover that already exists?

Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,228 @@ __int16 Abe::TryMountElum_42E600()
...
+        {
+            field_112_prev_motion = field_FE_next_state;
+        }
+
+        Abe_SFX_2_42A220(14, 0, 0x7FFF, this);
+        Abe_SFX_42A4D0(10, 127, 0, this);
+        Abe_SFX_2_42A220(7, 0, 0x7FFF, this);
+        SFX_Play_43AE60(SoundEffect::Unknown_79, 0, -500, this);
+        SFX_Play_43AD70(SoundEffect::KillEffect_78, 0, this);
+        return;
```
paulsapps, line 2481: redundant

Source/AliveLibAO/Abe.cpp
d3nwah: AE doesn't have it, do it for AE also?

Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,214 @@ __int16 Abe::TryMountElum_42E600()
...
+            {
+                field_106_shot = 0;
+                field_100_health = FP_FromInteger(1);
+                return;
+            }
+            if (shootKind != ShootKind::eEverythingElse_0)
+            {
+                if (shootKind == ShootKind::eHanging_1)
+                {
+                    field_FC_current_motion = 92;
```
paulsapps, line 2411: use the enum

Source/AliveLibAO/Abe.cpp
```diff
@@ -2258,6 +2260,214 @@ __int16 Abe::TryMountElum_42E600()
...
+                    field_108_bMotionChanged = 1;
+                    field_106_shot = 0;
+                    field_114_gnFrame = 0;
+                }
+                else if (shootKind == ShootKind::eRolling_2)
+                {
+                    field_FE_next_state = 147;
```
paulsapps, line 2418: enum

PR #651, opened 2020-09-28
impl some funcs

mouzedrift:


Source/AliveLibAO/BaseAnimatedWithPhysicsGameObject.cpp
```diff
@@ -47,9 +55,46 @@ BaseAnimatedWithPhysicsGameObject* BaseAnimatedWithPhysicsGameObject::ctor_417C1
     return this;
 }
 
-void BaseAnimatedWithPhysicsGameObject::Animation_Init_417FD0(int /*frameTableOffset*/, int /*maxW*/, int /*maxH*/, BYTE** /*ppAnimData*/, __int16 /*a6*/)
+void BaseAnimatedWithPhysicsGameObject::Animation_Init_417FD0(int frameTableOffset, int maxW, int maxH, BYTE** ppAnimData, __int16 a6)
 {
-    NOT_IMPLEMENTED();
+    const auto init = field_10_anim.Init_402D20(
```
paulsapps, line 60: might as well drop that var and put the call directly within in the if ?

Source/AliveLibAO/BaseAnimatedWithPhysicsGameObject.cpp
```diff
@@ -276,9 +321,20 @@ PSX_RECT* BaseAnimatedWithPhysicsGameObject::VGetBoundingRect_418120(PSX_RECT* p
...
 
-void BaseAnimatedWithPhysicsGameObject::SetTint_418750(const TintEntry* /*pTintArray*/, LevelIds /*levelId*/)
+void BaseAnimatedWithPhysicsGameObject::SetTint_418750(const TintEntry* pTintArray, LevelIds level_id)
 {
-    NOT_IMPLEMENTED();
+    while (pTintArray->field_0_level != static_cast<int>(level_id))
+    {
+        if (pTintArray->field_0_level == static_cast<int>(level_id) || pTintArray->field_0_level == static_cast<int>(LevelIds::eNone))
+        {
+            break;
```
paulsapps, line 330: is this OG? If it is and the item isn't found it seems like it could crash :oddtism:

Source/AliveLibAO/Particle.cpp
```diff
@@ -6,10 +6,32 @@
 
 START_NS_AO
 
-Particle* CC New_Particle_419D00(FP /*xpos*/, FP /*ypos*/, FP /*scale*/)
+Particle* CC New_Particle_DestroyOrCreateObject_419D00(FP xpos, FP ypos, FP scale)
```
paulsapps, line 9: rename the AE func to the same name?

Source/AliveLibAO/BaseAnimatedWithPhysicsGameObject.cpp
mouzedrift: just put it there because it reads a bit better imo /shrug

paulsapps: if you think its better then leave it as-is, its subjective :)

Source/AliveLibAO/BaseAnimatedWithPhysicsGameObject.cpp
mouzedrift: yup OG

PR #662, opened 2020-09-30
explosion and foot switch funcs

mouzedrift:


Source/AliveLibAO/FootSwitch.cpp
```diff
@@ -76,5 +77,54 @@ void FootSwitch::VScreenChanged_4889D0()
...
+    VUpdate_4888E0();
+}
+
+void FootSwitch::VUpdate_4888E0()
+{
+    switch (field_E8_state)
+    {
+    case States::eWaitForStepOnMe_0:
+        field_F0_pStoodOnMe = WhoIsStoodOnMe_488A60();
+        if (WhoIsStoodOnMe_488A60())
```
paulsapps, line 97: I think this should be using field_F0_pStoodOnMe  rather than calling it again

Source/AliveLibAO/FootSwitch.cpp
```diff
@@ -76,5 +77,54 @@ void FootSwitch::VScreenChanged_4889D0()
...
+        {
+            WhoIsStoodOnMe_488A60()->field_C_refCount++;
```
paulsapps, line 99: ditto

Issue #670, opened 2020-10-01
AO Rev Bug - Birds don't spawn


UltraStars3000:
**Description of issue:**
Pretty explicit uh?

**Steps to repro:**
Load the save: [bruh.zip](https://github.com/AliveTeam/alive_reversing/files/5312233/bruh.zip)

**Function:** `Factory::Factory_Dove_4834C0`

PR #672, opened 2020-10-01
#670 fix.

d3nwah:


Source/AliveLibAE/Factory.cpp
```diff
@@ -432,8 +432,8 @@ EXPORT void CC Factory_Dove_4D7E90(Path_TLV* pTlv, Path*, TlvItemInfoUnion tlvIn
             }
             else
             {
-                pDove->field_B8_xpos = FP_FromInteger((pTlv->field_8_top_left.field_0_x + width  * Math_NextRandom()) / 256);
-                ypos = (pTlv->field_8_top_left.field_2_y + height * Math_NextRandom()) / 256;
+                pDove->field_B8_xpos = FP_FromInteger(pTlv->field_8_top_left.field_0_x + width  * Math_NextRandom() / 256);
```
paulsapps, line 435: brackets should just be moved to the correct place rather than nuked

d3nwah: Personally, I don't see any place I deem "correct" for those, but feel free to add them if you think otherwise

paulsapps: I'll add them and hopefully not cause a regression

PR #675, opened 2020-10-01
[AO] Global SFX names Chapter 2

UltraStars3000:
Also deleted the whole logger to only display the 3 missing ones


Source/AliveLibAO/WinMain.cpp
```diff
@@ -581,6 +582,24 @@ EXPORT void CC Game_Loop_437630()
 
     while (!sBreakGameLoop_507B78 && !gBaseGameObject_list_9F2DF0->Empty())
     {
+        static unsigned char sfxId = 0;
```
paulsapps, line 585: can this get moved into its own func or removed if it isn't needed anymore?

PR #702, opened 2020-10-05
fix bird portal crash side effect

mouzedrift:


Source/AliveLibAO/Mudokon.cpp
```diff
@@ -1134,7 +1134,7 @@ __int16 Mudokon::FindBirdPortal_440250()
         return 0;
     }
 
-    if (FP_Abs(field_1AC_pBirdPortal->field_18_xpos - field_A8_xpos) < FP_FromInteger(gPsxDisplay_504C78.field_0_width) ||
+    if (FP_Abs(field_1AC_pBirdPortal->field_18_xpos - field_A8_xpos) < FP_FromInteger(gPsxDisplay_504C78.field_0_width << 16) ||
```
paulsapps, line 1137: FP_FromInteger is already doing the << 16 so now its << 32.... hmm

PR #724, opened 2020-10-08
impl Abe::State_35_RunLoop_425060

mouzedrift:


Source/AliveLibAO/Abe.cpp
```diff
@@ -5699,7 +5699,146 @@ void Abe::State_34_RunJumpLand_427560()
...
+
+    sub_422FC0();
+
+    if (field_FC_current_motion != eAbeStates::State_35_RunLoop_425060)
+    {
+        field_2A8_flags.Clear(Flags_2A8::e2A8_Bit3);
+        return;
+    }
+
+    if (field_10_anim.field_92_current_frame && field_10_anim.field_92_current_frame != 8)
```
paulsapps, line 5722: this looks suspect, should be != 0 I think?

Source/AliveLibAO/Abe.cpp
```diff
@@ -5699,7 +5699,146 @@ void Abe::State_34_RunJumpLand_427560()
...
+            field_2A8_flags.Clear(Flags_2A8::e2A8_Bit3);
+            field_10C_prev_held = 0;
+            return;
+        }
+    }
+    else
+    {
+        MapFollowMe_401D30(TRUE);
+
+        if (field_10C_prev_held & sInputKey_Hop_4C65A0)
```
paulsapps, line 5821: this whole block is repeated, you can just move it out of the else to avoid duplication and put a return before the else

Source/AliveLibAO/Abe.cpp
mouzedrift: nope it's actually supposed to be like that

paulsapps: != 0 is the same effect but more readable though

Source/AliveLibAO/Abe.cpp
```diff
@@ -5699,7 +5699,122 @@ void Abe::State_34_RunJumpLand_427560()
...
+                Environment_SFX_42A220(EnvironmentSfx::eRunSlide_4, 0, 0x7FFF, this);
+            }
+
+            field_2A8_flags.Clear(Flags_2A8::e2A8_Bit3);
+            field_10C_prev_held = 0;
+        }
+        return;
+    }
+
+    MapFollowMe_401D30(TRUE);
```
paulsapps, line 5797: wasn't that only done on one half of the if/else? 🤔

PR #759, opened 2020-10-12
Abe::DoGameSpeak_42F5C0() + style changes.

d3nwah:


Source/AliveLibAO/Abe.cpp
```diff
@@ -1746,11 +1724,123 @@ void Abe::ElumFree_4228F0()
...
+                }
+                New_Smoke_Particles_419A80(
+                    xPos,
+                    field_AC_ypos - (FP_FromInteger(24) * field_BC_sprite_scale),
+                    FP_FromDouble(0.5) * field_BC_sprite_scale,
+                    3,
+                    1
+                );
+            }
+            field_130_say = 8;
```
paulsapps, line 1834: wasn't there an enum for this?

mouzedrift: there is: ``MudSounds``

d3nwah: can't be, wrong type

PR #781, opened 2020-10-19
Abe::State_88_HandstoneBegin_430590()

d3nwah:
✋🗿

### DIFF
--- a/Source/AliveLibAO/Abe.cpp
+++ b/Source/AliveLibAO/Abe.cpp
@@ -7909,14 +7909,10 @@ void Abe::State_88_HandstoneBegin_430590()
                     0
                 );
 
-                if (field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX))
-                {
-                    field_164_pCircularFade->field_10_anim.field_4_flags.Set(AnimFlags::eBit5_FlipX);
-                }
-                else
-                {
-                    field_164_pCircularFade->field_10_anim.field_4_flags.Clear(AnimFlags::eBit5_FlipX);
-                }
+                field_164_pCircularFade->field_10_anim.field_4_flags.Set(
+                    AnimFlags::eBit5_FlipX,
+                    field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX)
+                );
 
                 field_110_state = StoneStates::eUnknown_1;
                 SFX_Play_43AD70(SoundEffect::IngameTransition_107, 90, 0);
@@ -8060,7 +8056,7 @@ void Abe::State_88_HandstoneBegin_430590()
         }
         case StoneStates::eUnknown_2:
         {
-            if (!sMovie_ref_count_9F309C)
+            if (sMovie_ref_count_9F309C == 0)
             {
                 gPsxDisplay_504C78.PutCurrentDispEnv_40DE40();
                 pScreenManager_4FF7C8->DecompressCameraToVRam_407110(
@@ -8080,7 +8076,7 @@ void Abe::State_88_HandstoneBegin_430590()
             {
                 field_110_state = StoneStates::eUnknown_5;
                 ResourceManager::FreeResource_455550(gpDemoPlaybackRes_50772C);
-                auto pCircularFade = Make_Circular_Fade_447640(
+                field_164_pCircularFade = Make_Circular_Fade_447640(
                     field_A8_xpos,
                     field_AC_ypos,
                     field_BC_sprite_scale,
@@ -8088,15 +8084,10 @@ void Abe::State_88_HandstoneBegin_430590()
                     0
                 );
 
-                field_164_pCircularFade = pCircularFade;
-                if (field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX))
-                {
-                    field_164_pCircularFade->field_10_anim.field_4_flags.Set(AnimFlags::eBit5_FlipX);
-                }
-                else
-                {
-                    field_164_pCircularFade->field_10_anim.field_4_flags.Clear(AnimFlags::eBit5_FlipX);
-                }
+                field_164_pCircularFade->field_10_anim.field_4_flags.Set(
+                    AnimFlags::eBit5_FlipX,
+                    field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX)
+                );
             }
             break;
         }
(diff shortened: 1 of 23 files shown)


Source/AliveLibAO/Abe.cpp
```diff
@@ -628,7 +632,7 @@ Abe* Abe::ctor_420770(int frameTableOffset, int /*r*/, int /*g*/, int /*b*/)
 
     field_10C_prev_held = 0; // lowest to base class
     field_E6_last_anim_frame = 0;
-    field_110_state = 0;
+    field_110_state = StoneStates::eUnknown_0;
```
paulsapps, line 635: this isn't just stone states so it should prob only use this enum in the stone state machine func...

Source/AliveLibAO/Abe.cpp
```diff
@@ -7885,9 +7887,322 @@ void Abe::State_87_428FA0()
     State_3_Fall_42E7F0();
 }
 
+ALIVE_VAR(1, 0x507728, int, gConter_507728, 0);
+ALIVE_VAR(1, 0x50772C, BYTE**, gpDemoPlaybackRes_50772C, 0);
```
paulsapps, line 7891: nullptr

Source/AliveLibAO/Abe.cpp
```diff
@@ -7885,9 +7887,322 @@ void Abe::State_87_428FA0()
...
+                    field_A8_xpos,
+                    field_AC_ypos,
+                    field_BC_sprite_scale,
+                    1,
+                    0
+                );
+
+                if (field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX))
+                {
+                    field_164_pCircularFade->field_10_anim.field_4_flags.Set(AnimFlags::eBit5_FlipX);
```
paulsapps, line 7914: you could just pass the .Get() into the 2nd arg of set instead, doesn't matter though

Source/AliveLibAO/Abe.cpp
```diff
@@ -7885,9 +7887,322 @@ void Abe::State_87_428FA0()
...
+                            FP_GetExponent(field_A8_xpos),
+                            FP_GetExponent(field_AC_ypos),
+                            TlvTypes::HandStone_100
+                        );
+                }
+
+                auto ptlv = static_cast<Path_Stone*>(field_F0_pTlv);
+                if (ptlv)
+                {
+                    field_174_pathStone = ptlv->field_18_data;
```
paulsapps, line 7961: the OG reads memory out of bounds here, we should probably fix it by switching on field_F0_pTlv->type and copying the corresponding sub struct of the union?

Source/AliveLibAO/Abe.cpp
```diff
@@ -7885,9 +7887,322 @@ void Abe::State_87_428FA0()
...
+                    }
+                    default:
+                        return;
+                }
+            }
+            break;
+        }
+        case StoneStates::eUnknown_2:
+        {
+            if (!sMovie_ref_count_9F309C)
```
paulsapps, line 8063: == 0 ?

Source/AliveLibAO/Abe.cpp
```diff
@@ -7885,9 +7887,322 @@ void Abe::State_87_428FA0()
...
+                pScreenManager_4FF7C8->field_36_flags |= 1;
+                field_164_pCircularFade->Vsub_479FE0(0, 0);
+                field_110_state = StoneStates::eUnknown_5;
+            }
+            break;
+        }
+        case StoneStates::eUnknown_3:
+        {
+            gConter_507728--;
+            if (!gConter_507728)
```
paulsapps, line 8079: == 0? also typo of counter

Source/AliveLibAO/Abe.cpp
```diff
@@ -7885,9 +7887,322 @@ void Abe::State_87_428FA0()
...
+                    field_AC_ypos,
+                    field_BC_sprite_scale,
+                    0,
+                    0
+                );
+
+                field_164_pCircularFade = pCircularFade;
+                if (field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX))
+                {
+                    field_164_pCircularFade->field_10_anim.field_4_flags.Set(AnimFlags::eBit5_FlipX);
```
paulsapps, line 8094: could maybe use Get() in the 2nd arg of Set here also

Source/AliveLibAO/Abe.cpp
d3nwah: Maybe rename to something akin to `InternalStates`?

Source/AliveLibAO/Abe.cpp
```diff
@@ -7885,9 +7887,322 @@ void Abe::State_87_428FA0()
...
+                    SFX_Play_43AD70(107u, 90, 0);
+                }
+            }
+            break;
+        }
+        case StoneStates::eUnknown_7:
+        {
+            if (field_158_pDeathFadeout->field_6E_bDone)
+            {
+                auto camera = field_174_pathStone.dataHandstone.cameras[field_16E_cameraIdx];
```
paulsapps, line 8151: could be const? maybe use the actual type here as well since its not obvious what it is

Source/AliveLibAO/Abe.cpp
```diff
@@ -7885,9 +7887,322 @@ void Abe::State_87_428FA0()
...
+        case StoneStates::eUnknown_13:
+        {
+            field_158_pDeathFadeout->field_6_flags.Set(Options::eDead_Bit3);
+            field_158_pDeathFadeout = 0;
+
+            field_164_pCircularFade = Make_Circular_Fade_447640(field_A8_xpos, field_AC_ypos, field_BC_sprite_scale, 0, 0);
+            field_110_state = StoneStates::eUnknown_5;
+            if (field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX))
+            {
+                field_164_pCircularFade->field_10_anim.field_4_flags.Set(AnimFlags::eBit5_FlipX);
```
paulsapps, line 8189: same get/set pattern

Source/AliveLibAO/DemoPlayback.cpp
```diff
@@ -0,0 +1,13 @@
+#include "stdafx_ao.h"
+#include "Function.hpp"
+#include "DemoPlayback.hpp"
+
+START_NS_AO
+
+EXPORT DemoPlayback* DemoPlayback::ctor_4517B0(BYTE**, __int16)
+{
+    NOT_IMPLEMENTED();
```
paulsapps, line 9: return this else build rip

Issue #783, opened 2020-10-19
AO Rev Bug - Slig possessing edge case behaviour inconsistent with OG


d3nwah:
Steps to repro (example):
- Start the game
- Possess the slig on the second screen being on the ledge above
- DDcheat fly the slig on the ledge
- Shoot abe so he falls down and doesn't die (Not the reason for making this ticket)
- Slig is still controllable, while he's not in OG (The reason for making this ticket)

Issue #777, opened 2020-10-17
AO Rev Bug - Bullet shells don't work as expected


mouzedrift:
OG | REV
--- | ---
![3lsBIo9h74](https://user-images.githubusercontent.com/43358824/96347697-b9ed9180-10a3-11eb-84e7-ccf982c3f97f.gif) | ![lgtz9dLdBe](https://user-images.githubusercontent.com/43358824/96347702-c1ad3600-10a3-11eb-9ee5-7150813328f3.gif)

Steps to repro:
- locate the slig above and possess him
- shoot
- 👀 

func:
BulletShell::VUpdate_4628C0

Issue #770, opened 2020-10-14
AO Rev Bug - Well doesn't work


mouzedrift:
Steps to repro:
- load the save
- enter the screen to your right
- jump into the well

save:
[!ewell.zip](https://github.com/AliveTeam/alive_reversing/files/5374914/ewell.zip)

func:
Abe::State_77_WellBegin_430F10

Issue #763, opened 2020-10-13
AO Rev Bug - Error when loading a specific save


mouzedrift:
Steps to repro:
- load the save

save:
[!a.zip](https://github.com/AliveTeam/alive_reversing/files/5372760/a.zip)

func:
SaveGame::Read_459D30

PR #786, opened 2020-10-20
#770 fix, #763 fix, #777 fix, #783 fix, #698 fix, #758 fix

d3nwah:


Source/AliveLibAO/Abe.cpp
```diff
@@ -7550,8 +7550,8 @@ void Abe::State_77_WellBegin_430F10()
     if (field_10_anim.field_4_flags.Get(AnimFlags::eBit18_IsLastFrame))
     {
         field_114_gnFrame = 15;
-        SFX_Play_43AD70(25u, 0, this);
-        field_FC_current_motion = eAbeStates::State_78_InsideWellLocal_4310A0;
+        SFX_Play_43AD70(SoundEffect::WellEnter_25, 0, this);
+        field_FC_current_motion = static_cast<eAbeStates>((static_cast<int>(field_FC_current_motion) + 1));
```
paulsapps, line 7554: cast is redundant because it isn't strongly type, but eventually we want this to be strongly typed so I think you should do something like the other states do:

```
switch(field_FC_current_motion)
{
case state_77_WellBegin_430F10:
 field_FC_current_motion = State_78;
 break;

// Whatever at least the other case is

default:
LOG_ERROR(field_FC_current_motion );
ALIVE_FATAL("Unhandled next well state);
break;
}

```

Source/AliveLibAO/Abe.cpp
```diff
@@ -7550,8 +7552,28 @@ void Abe::State_77_WellBegin_430F10()
...
+                field_FC_current_motion = eAbeStates::State_81_InsideWellExpress_431320;
+                break;
+            }
+            case eAbeStates::State_83_430F00:
+            {
+                field_FC_current_motion = eAbeStates::State_84_431080;
+                break;
+            }
+            default:
+                ALIVE_FATAL("Unrecognized stat called Abe::State_77_WellBegin_430F10!");
```
paulsapps, line 7575: maybe log so if it gets reported we know which one to add :shook:

Source/AliveLibAO/Abe.cpp
```diff
@@ -3064,6 +3066,8 @@ __int16 Abe::VTakeDamage_4214E0(BaseGameObject* pFrom)
 
     field_11C_regen_health_timer = gnFrameCount_507670 + 900;
 
+    auto oldHpGtZero = field_100_health > FP_FromInteger(0);
```
paulsapps, line 3069: surely just store const FP oldHP = field_100_health ; and check > 0 at the end? :P

d3nwah: it's not const on purpose :shook:

paulsapps: I mean about checking > 0 in the local var - that is an odd compiler/ida-ism

PR #789, opened 2020-10-20
Abe::State_156_DoorEnter_42D370()

d3nwah:


Source/AliveLibAO/Abe.cpp
```diff
@@ -9535,9 +9535,197 @@ void Abe::State_155_ElumJumpToFall_42E060()
...
+                field_118 = gnFrameCount_507670 + 3;
+            }
+            return;
+        }
+        case DoorStates::eUnknown_3:
+        {
+            if (field_118 <= static_cast<int>(gnFrameCount_507670))
+            {
+                field_110_state.door = DoorStates::eUnknown_4;
+                auto pTlv = static_cast<Path_Reset*>(gMap_507BA8.TLV_Get_At_446260(
```
paulsapps, line 9586: assign to field_F0_pTlv  missing?

Source/AliveLibAO/Abe.cpp
```diff
@@ -9535,9 +9535,197 @@ void Abe::State_155_ElumJumpToFall_42E060()
...
+                field_A8_xpos,
+                FP_FromInteger(field_F0_pTlv->field_10_top_left.field_2_y),
+                field_A8_xpos,
+                FP_FromInteger(field_F0_pTlv->field_14_bottom_right.field_2_y),
+                &pLine,
+                &hitX,
+                &hitY,
+                field_BC_sprite_scale != FP_FromDouble(0.5) ? 7 : 0x70))
+            {
+                field_F4_pLine = pLine;
```
paulsapps, line 9705: might as well use field_F4_pLine  directly?

Source/AliveLibAO/Abe.cpp
```diff
@@ -9535,9 +9535,197 @@ void Abe::State_155_ElumJumpToFall_42E060()
...
+        {
+            field_B2_lvl_number = gMap_507BA8.field_0_current_level;
+            field_B0_path_number = gMap_507BA8.field_2_current_path;
+            gMap_507BA8.field_1E_door = 0;
+            auto pPathDoor = static_cast<Path_Door*>(gMap_507BA8.TLV_First_Of_Type_In_Camera_4464A0(Door_6, 0));
+            field_F0_pTlv = pPathDoor;
+
+            while (pPathDoor->field_20_door_number != field_196_door_id)
+            {
+                pPathDoor = static_cast<Path_Door*>(Path_TLV::TLV_Next_Of_Type_446500(
```
paulsapps, line 9652: this is wrong because you've missed checking a door 

get first (don't get check it)
get next (always checked from here on)

paulsapps: actually I think that is correct due to the while - but should field_F0_pTlv be assigned even when it isn't the door we are looking for? 🤔

Source/AliveLibAO/Abe.cpp
d3nwah: not this time, no. At least I can't see it in IDA as well

Source/AliveLibAO/Abe.cpp
d3nwah: Pushed, but I'm not sure if I like doing it so implicitly...

Source/AliveLibAO/Abe.cpp
d3nwah: I think the code's alright as is

Source/AliveLibAO/Map.cpp
```diff
@@ -1442,9 +1442,45 @@ Path_TLV* Map::TLV_Get_At_446060(Path_TLV* pTlv, FP xpos, FP ypos, FP width, FP
...
+
+    if (counterInit > 0)
+    {
+        DWORD* pObjectTable = reinterpret_cast<DWORD*>(&pPathRes[pPathData->field_18_object_index_table_offset]);
+
+        for(auto counter = 0; counter < counterInit; counter++)
+        {
+            if (pObjectTable[counter] != -1)
+            {
+                auto pTlv = (Path_TLV *) &pPathRes[pPathData->field_14_object_offset + pObjectTable[counter]];
```
paulsapps, line 1462: reinterpret_cast

Source/AliveLibAO/Map.cpp
```diff
@@ -1442,9 +1442,45 @@ Path_TLV* Map::TLV_Get_At_446060(Path_TLV* pTlv, FP xpos, FP ypos, FP width, FP
     return pTlv;
 }
 
-void Map::sub_447430(unsigned __int16)
+void Map::sub_447430(unsigned __int16 pathNum)
 {
-    NOT_IMPLEMENTED();
+    const auto pPathData = Path_Get_Bly_Record_434650(field_0_current_level, pathNum)->field_4_pPathData;
+    auto pPathRes = *field_5C_path_res_array.field_0_pPathRecs[pathNum];
```
paulsapps, line 1448: const

Source/AliveLibAO/Map.cpp
```diff
@@ -1442,9 +1442,45 @@ Path_TLV* Map::TLV_Get_At_446060(Path_TLV* pTlv, FP xpos, FP ypos, FP width, FP
...
+
+    auto counterInit = (pPathData->field_A_bBottom - pPathData->field_6_bRight)
```
paulsapps, line 1450: const

PR #804, opened 2020-10-27
Slig::HandlePlayerControlled_4667B0()  Slig::Slig_GameSpeak_SFX_46F560()

d3nwah:


Source/AliveLibAO/Slig.cpp
```diff
@@ -1763,9 +1766,21 @@ __int16 CCSTD Slig::IsWallBetween_46BE60(Slig* pLeft, BaseAliveGameObject* pRigh
...
+        volume = sSligSounds_4CFB30[static_cast<int>(effectId) + 6].field_C_default_volume;
+    }
+    if (pObj)
+    {
+        if (pObj->field_BC_sprite_scale == FP_FromDouble(0.5))
+        {
+            volume = FP_GetExponent(FP_FromInteger(volume * 2) / FP_FromInteger(3));
+        }
+    }
+    SFX_SfxDefinition_Play_4770F0(&sSligSounds_4CFB30[static_cast<int>(effectId) + 6], volume, pitch_min, pitch_min);
```
paulsapps, line 1783: static_cast<int>(effectId) + 6 should really be fixed - its mapping from SligSpeak to SoundEffects enum I believe, I hope AE hasn't got this hack as well

Source/AliveLibAO/Slig.cpp
```diff
@@ -2069,9 +2084,253 @@ void Slig::MoveOnLine_467490()
...
 }
 
+void Slig::OperateLift()
+{
+    if (field_10_anim.field_4_flags.Get(AnimFlags::eBit5_FlipX))
+    {
+        field_FC_current_motion = eSligStates::State_5_TurnAroundStanding_469C80;
+        field_FE_next_state = eSligStates::State_49_LiftGrip_4663A0;
+
+        auto camPos = gMap_507BA8.GetDirection(field_B2_lvl_number, field_B0_path_number, field_A8_xpos, field_AC_ypos);
```
paulsapps, line 2094: const+ maybe type it

Source/AliveLibAO/Slig.cpp
```diff
@@ -2069,9 +2084,253 @@ void Slig::MoveOnLine_467490()
...
+    }
+}
+
 signed __int16 Slig::HandlePlayerControlled_4667B0()
 {
-    NOT_IMPLEMENTED();
+    FP kScaleGrid = ScaleToGridSize_41FA30(field_BC_sprite_scale);
+
+    auto pressed = sInputObject_5009E8.field_0_pads[sCurrentControllerIndex_5076B8].field_0_pressed;
+    if (pressed & sInputKey_Right_4C6590)
```
paulsapps, line 2118: const or use sInputObject_5009E8.isPressed()

Source/AliveLibAO/Slig.cpp
```diff
@@ -2069,9 +2084,253 @@ void Slig::MoveOnLine_467490()
...
+        {
+            field_FC_current_motion = eSligStates::State_5_TurnAroundStanding_469C80;
+            return 1;
+        }
+    }
+
+    if (sInputObject_5009E8.isHeld(sInputKey_DoAction_4C65A4)
+        && !(pressed & (sInputKey_LeftGameSpeakEnabler_4C65B8 | sInputKey_RightGameSpeakEnabler_4C65DC)))
+    {
+        if (Input_IsChanting_4334C0() == FALSE)
```
paulsapps, line 2191: I think everything else is using !Input_IsChanting_4334C0()

d3nwah: Shouldn't we use this as it's more explicit, and this function doesn't return proper `bool`? (yet)

paulsapps: shouldn't matter as !bool !int !short all convert to a boolean expression so it would need less changes for int -> bool this way

Source/AliveLibAO/Slig.cpp
d3nwah: const

PR #833, opened 2020-11-05
name some flags, funcs, add an InvisibleState enum and fix broken foot switch

mouzedrift:


Source/AliveLibAE/FootSwitch.hpp
```diff
@@ -8,7 +8,7 @@ enum class SwitchOp : __int16;
 
 enum class FootSwitchTriggerBy : __int16
 {
-    eOnlyAbe_0 = 1,
+    eOnlyAbe_0 = 0,
```
paulsapps, line 11: ahhhhhhhhhhh

mouzedrift: tapeDan

PR #837, opened 2020-11-07
[AE] Named Almost Everything in the Abe Class

THEONLYDarkShadow:
Named fields, flags, structures and functions in the `Abe` class. Also named an object type in the `BaseGameObject` class and a couple of fields in the `MusicController` class.

### DIFF
--- a/Source/AliveLibAE/Abe.hpp
+++ b/Source/AliveLibAE/Abe.hpp
@@ -226,22 +226,22 @@ struct Abe_1BC_20_sub_object
 {
     int field_0_gnFrame;
     int field_4_regen_health_timer;
-    FP field_8;
-    int field_C;
-    unsigned __int16 field_10;
+    FP field_8_x_vel_slow_by;
+    int field_C_unused;
+    unsigned __int16 field_10_resource_index;
     Mud_Emotion field_12_mood;
-    int field_14;
+    int field_14_rolling_motion_timer;
     MudSounds field_18_say;
-    __int16 field_1A;
+    __int16 field_1A_padding;
 };
 ALIVE_ASSERT_SIZEOF(Abe_1BC_20_sub_object, 0x1C);
 
-EXPORT void CC Mudokon_SFX_457EC0(MudSounds idx, __int16 volume, int pitch, BaseAliveGameObject *pHero);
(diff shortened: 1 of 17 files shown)


Source/AliveLibAE/Abe.hpp
```diff
@@ -300,58 +300,58 @@ struct Abe_SaveState
...
+        eD4_Bit9_unused = 0x100,
+        eD4_Bit10_play_ledge_grab_sounds = 0x200,
+        eD4_Bit11_unused = 0x400,
+        eD4_Bit12_have_healing = 0x800,
         eD4_eBit13 = 0x1000,
-        eD4_eBit14 = 0x2000,
-        eD4_eBit15 = 0x4000,
-        eD4_eBit16 = 0x8000,
+        eD4_eBit14_is_mudanchee_vault_ender = 0x2000,
+        eD4_eBit15_is_mudanchee_vault_ender = 0x4000,
```
paulsapps, line 345: 14 and 15 are the same? 🤔

Source/AliveLibAE/Abe.hpp
```diff
@@ -300,58 +300,58 @@ struct Abe_SaveState
...
+        eD4_Bit4_unused = 0x8,
+        eD4_Bit5_prevent_chanting = 0x10,
+        eD4_Bit6_land_softly = 0x20,
+        eD4_Bit7_unused = 0x40,
+        eD4_Bit8_laugh_at_chant_end = 0x80,
+        eD4_Bit9_unused = 0x100,
+        eD4_Bit10_play_ledge_grab_sounds = 0x200,
+        eD4_Bit11_unused = 0x400,
+        eD4_Bit12_have_healing = 0x800,
         eD4_eBit13 = 0x1000,
```
paulsapps, line 343: bit 13 not named?

THEONLYDarkShadow: It connects to `Flags_114::e114_Bit10` in the `BaseAliveGameObject` class, so I'll tackle that one later.

Source/AliveLibAE/Abe.hpp
THEONLYDarkShadow: For some reason, I thought they were supposed to be the same. When I looked at the code again, I realised that one is supposed to be `eD4_eBit15_is_mudomo_vault_ender`. :sweat_smile: The fix has been pushed now.

PR #855, opened 2020-11-08
Fix rounding errors, improve rendering on high DPI displays

LIJI32:
This commit fixes rounding error when maintaining aspect ratio, and improves rendering on Retina Macs by enabling high DPI support in SDL.

### DIFF
--- a/Source/AliveLibAE/VGA.cpp
+++ b/Source/AliveLibAE/VGA.cpp
@@ -156,20 +156,20 @@ EXPORT void CC VGA_CopyToFront_4F3730(Bitmap* pBmp, RECT* pRect, int /*screenMod
 
             int w = 0;
             int h = 0;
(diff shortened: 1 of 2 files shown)


Source/AliveLibAE/VGA.cpp
```diff
@@ -156,20 +156,20 @@ EXPORT void CC VGA_CopyToFront_4F3730(Bitmap* pBmp, RECT* pRect, int /*screenMod
 
             int w = 0;
             int h = 0;
-            SDL_GetWindowSize(Sys_GetHWnd_4F2C70(), &w, &h);
+            SDL_GL_GetDrawableSize(Sys_GetHWnd_4F2C70(), &w, &h);
```
paulsapps, line 159: we're not using GL to render so I'm not sure this is correct? GL might get picked as the backend but it could be something else...

LIJI32:
True, switched to a more portable API.

PR #862, opened 2020-11-11
Rope::VRender_458780() Spark::VUpdate_477DB0() DoorFlame::VUpdate_432440()

d3nwah:


Source/AliveLibAO/DoorFlame.cpp
```diff
@@ -238,7 +239,47 @@ class FlameSparks : public BaseAnimatedWithPhysicsGameObject
 
     EXPORT void VUpdate_432440()
     {
-        NOT_IMPLEMENTED();
+        PSX_RECT rect = {};
+        gMap_507BA8.Get_Camera_World_Rect_444C30(CameraPos::eCamCurrent_0, &rect);
+        field_A8_xpos = FP_FromInteger(rect.w + 16);
+        field_AC_ypos = FP_FromInteger(rect.y + 0xFFF0);
```
paulsapps, line 245: is that not -16 or something? 🤔

Source/AliveLibAO/Rope.cpp
```diff
@@ -142,9 +145,109 @@ void Rope::VRender(int** pOrderingTable)
     VRender_458780(pOrderingTable);
 }
 
-void Rope::VRender_458780(int** /*ppOt*/)
+void Rope::VRender_458780(int** ppOt)
 {
-    NOT_IMPLEMENTED();
+    PSX_Point camPos;
```
paulsapps, line 150: = {}

Source/AliveLibAO/Rope.cpp
```diff
@@ -142,9 +145,109 @@ void Rope::VRender(int** pOrderingTable)
...
+        if (field_B0_path_number == gMap_507BA8.field_2_current_path)
+        {
+            if (field_A8_xpos >= FP_FromInteger(camPos.field_0_x) &&
+                field_A8_xpos <= FP_FromInteger(camPos.field_0_x + 1024))
+            {
+                const FP camXPos = pScreenManager_4FF7C8->field_10_pCamPos->field_0_x;
+                const FP camYPos = pScreenManager_4FF7C8->field_10_pCamPos->field_4_y;
+
+
+                int minY = FP_GetExponent((FP_FromInteger(pScreenManager_4FF7C8->field_16_ypos + (unsigned __int16) field_EE_top))
```
paulsapps, line 163: casts seem redundant, if not then static_cast

Source/AliveLibAO/Rope.cpp
```diff
@@ -142,9 +145,109 @@ void Rope::VRender(int** pOrderingTable)
...
+                        field_E8_pRopeRes[idx].field_9_g = static_cast<BYTE>(g);
+                        field_E8_pRopeRes[idx].field_A_b = static_cast<BYTE>(b);
+
+                        field_E8_pRopeRes[idx].VRender2(
+                            screenX,
+                            field_C8_yOffset + screenY,
+                            ppOt
+                        );
+
+                        PSX_RECT pOut = {};
```
paulsapps, line 225: its called pOut but isn't a pointer, isn't it just frameRect ?

Source/AliveLibAO/DoorFlame.cpp
d3nwah: Nope, not sure what that is, because if it was -16, the number would be twice as long (rect.y is short) - in AE it *is* twice as long in IDA...

PR #886, opened 2020-11-21
name fields, enums and fix some bugs

mouzedrift:


mouzedrift:
can you not fail again sdl2 plz :oddtism:

Source/AliveLibAO/Grenade.cpp
```diff
@@ -427,7 +427,7 @@ signed __int16 Grenade::InTheAir_41EF10()
         &field_114_pCollisionLine,
         &hitX,
         &hitY,
-        field_BC_sprite_scale != FP_FromInteger(1) ? 0x70 : 0x07);
+        field_BC_sprite_scale != FP_FromDouble(0.5) ? 0xFFFFFFF1 : 0);
```
paulsapps, line 430: I think that is just 1 rather than 0xFFFFFFF1 since its a bit index

PR #888, opened 2020-11-21
Impl meat funcs

mouzedrift:


Source/AliveLibAO/Meat.cpp
```diff
@@ -346,18 +463,206 @@ void Meat::VUpdate()
...
+            {
+                BaseGameObject* pObjIter = ObjListPlatforms_50766C->ItemAt(i);
+                if (!pObjIter)
+                {
+                    break;
+                }
+
+                if (pObjIter->field_4_typeId == Types::eLiftPoint_51 || pObjIter->field_4_typeId == Types::eTrapDoor_98)
+                {
+                    auto pIteratedLiftPoint = static_cast<LiftPoint*>(pObjIter);
```
paulsapps, line 637: should probabably cast to PlatformBase* since it can be a lift point or a trap door

PR #927, opened 2020-12-01
OG memory corruption fix for Free_Resources_For_Camera_44717

paulsapps:


paulsapps:
this is what AE does so gonna merge it 👻

Issue #838, opened 2020-11-07
AO Rev Bug - Glitching graphics (bad rect)


mouzedrift:
Steps to repro:
- load the save
- go to the right until you find a lever and pull it
- go back and enter the well to scrabania

save:
[!2.zip](https://github.com/AliveTeam/alive_reversing/files/5504251/2.zip)

func:
Animation::Init_402D20

pics:
![AbeWin_AliveDllAO_p5TqEuU113](https://user-images.githubusercontent.com/43358824/98431258-fa916700-20b3-11eb-8318-c700ec8bac7f.png)  ![image](https://user-images.githubusercontent.com/43358824/98431295-3e846c00-20b4-11eb-837a-a1d178e3617d.png)

PR #940, opened 2020-12-03
#838 fix.

d3nwah:


Source/AliveLibAO/Animation.cpp
```diff
@@ -907,7 +911,8 @@ signed __int16 Animation::Init_402D20(int frameTableOffset, DynamicArray* /*anim
     }
 
     // NOTE: OG bug or odd compiler code gen? Why isn't it using the passed in list which appears to always be this anyway ??
-    if (!gObjList_animations_505564->Push_Back(this))
+    const auto result = gObjList_animations_505564->Push_Back(this);
+    if (!result)
     {
         return 0;
```
paulsapps, line 917: really this should never happen, perhaps we should log an error here (maybe it will appear when stuff vanishes)

PR #978, opened 2020-12-08
Fix #962

### LINKED ISSUE #962: AO Rev Bug - Font is slightly misaligned
REV | OG
--- | ---
AbeWin_AliveDllAO_RQcFrcnirA | AbeWin_AliveDllAO_taGglhMtvR

func:
Font::MeasureWidth_41C2B0

FrozenFish24:
Fixes the slightly skewed button prompt text and misaligned save names.

There are several other uses of `result = static_cast<int>(result * 0.575)` in `Font.cpp`  that may need to be re-evaluated.

### DIFF
--- a/Source/AliveLibAO/Font.cpp
+++ b/Source/AliveLibAO/Font.cpp
@@ -480,7 +480,7 @@ EXPORT int AliveFont::DrawString_41C360(int** ot, const char* text, __int16 x, _
 
         ++characterRenderCount;
 
-        offsetX += widthScaled + static_cast<short>(field_34_font_context->field_8_atlas_array[0].field_2_width * FP_GetDouble(scale));
+        offsetX += widthScaled + FP_GetExponent(FP_FromInteger(field_34_font_context->field_8_atlas_array[0].field_2_width) * scale);
 
         poly += 2;
     }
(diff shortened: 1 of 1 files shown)


Source/AliveLibAO/Font.cpp
```diff
@@ -479,7 +480,7 @@ EXPORT int AliveFont::DrawString_41C360(int** ot, const char* text, __int16 x, _
 
         ++characterRenderCount;
 
-        offsetX += widthScaled + static_cast<short>(field_34_font_context->field_8_atlas_array[0].field_2_width * FP_GetExponent(scale));
+        offsetX += widthScaled + static_cast<short>(field_34_font_context->field_8_atlas_array[0].field_2_width * FP_GetDouble(scale));
```
paulsapps, line 483: FP_GetDouble was supposed to be for debug printing of FixedPoints as float/double. Probably you want to convert width to a fixed point and then convert that fixed point to an integer, something like:

offsetX += widthScaled + FP_GetExponent(FP_FromInteger(field_34_font_context->field_8_atlas_array[0].field_2_width) * scale);

Assuming scale is an FP.

FrozenFish24: Can do

PR #990, opened 2020-12-11
[AE] MainMenu class refactoring, MainMenuCams enum

pryon:
Some Update function return values are still in int format (0xFFFF0002, etc.).
As we've discussed, those will be converted to a struct, but that will have to be in another PR as this one has already gotten too big

### DIFF
--- a/Source/AliveLibAE/MainMenu.cpp
+++ b/Source/AliveLibAE/MainMenu.cpp
@@ -158,65 +158,72 @@ SfxDefinition stru_55D7C0[] =
 
 ALIVE_ARY(1, 0x561960, MainMenuPage, 24, sMainMenuPages_561960, 
 {
-    { // 0
-        23,       0,        0,        0,        -1,        0,        0,
+    { // Page 0: Controller selection menu
+        MainMenuCams::eControllerSelectionCam,
+        0,        0,        0,       -1,        0,        0,
         &MainMenuController::ControllerMenu_Update_4D16D0,
         &MainMenuController::ControllerMenu_Render_Text_4D26C0,
         nullptr,
         &MainMenuController::ControllerMenu_Load_4D16B0,
         nullptr
     }, 
-    { // 1
-        5,        0,        0,        0,        -1,        0,        0,
+    { // Page 1: Input remapping menu
+        MainMenuCams::eRemapInputsCam,
+        0,        0,        0,       -1,        0,        0,
         &MainMenuController::RemapInput_Update_4D1820,
         &MainMenuController::RemapInput_Render_4D2A10,
         nullptr,
         &MainMenuController::RemapInput_Load_4D17E0,
         nullptr
     }, 
-    { // 2
-        11,        0,        0,        0,        -1,        0,       0,
+    { // Page 2: Dummy blank screen for quitting
+        MainMenuCams::eDummyBlankCam,
+        0,        0,        0,       -1,        0,        0,
         nullptr,
         nullptr,
         nullptr,
         &MainMenuController::Game_Force_Quit_Load_4D1A90,
         nullptr
     },
-    { // 3
-        // Main menu screen where abe says hello
-        1,        0,        0,        0,        -1,        -1,    1,
+    { // Page 3: Main menu screen where abe says hello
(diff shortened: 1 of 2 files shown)


Source/AliveLibAE/MainMenu.cpp
```diff
@@ -2408,39 +2435,80 @@ signed int MainMenuController::AbeMotions_Update_4D1F50(DWORD input)
...
+    return MainMenuCams::eNoChange;
+}
 
-signed int MainMenuController::tsub_4D48C0(DWORD /*input*/)
+signed int MainMenuController::PSX_Gamemode_Selection_Update_4D48C0(DWORD input)
 {
-    NOT_IMPLEMENTED();
-    return 0;
+    __int16 result; // ax
+    BitField32<Flags> menuFlags = this->field_23C_T80;
```
paulsapps, line 2474: you can use field_23C_T80 directly instead of a local

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -2408,39 +2435,80 @@ signed int MainMenuController::AbeMotions_Update_4D1F50(DWORD input)
...
     }
 }
 
-signed int MainMenuController::tsub_Input_Check_4D49B0(DWORD /*input*/)
+signed int MainMenuController::PSX_Cooperative_Mode_Update_4D49B0(DWORD /*input*/)
 {
-    NOT_IMPLEMENTED();
-    return 0;
-}
+    DWORD held = sInputObject_5BD4E0.field_0_pads[0].field_C_held;
```
paulsapps, line 2462: can be const

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -2408,39 +2435,80 @@ signed int MainMenuController::AbeMotions_Update_4D1F50(DWORD input)
...
-    NOT_IMPLEMENTED();
-    return 0;
+    __int16 result; // ax
+    BitField32<Flags> menuFlags = this->field_23C_T80;
+
+    if (input & InputCommands::eUnPause_OrConfirm)
+    {
+        sGameStartedFrame_5C1B88 = sGnFrame_5C1B84;
+        sCurrentControllerIndex_5C1BBE = 0;
+        BOOL twoPlayerModeSelected = this->field_1FC_button_index == 1;
```
paulsapps, line 2480: can be const and this-> is redundant

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -2408,39 +2435,80 @@ signed int MainMenuController::AbeMotions_Update_4D1F50(DWORD input)
...
 
+    return MainMenuCams::eNoChange;
+}
 
-signed int MainMenuController::tsub_4D48C0(DWORD /*input*/)
+signed int MainMenuController::PSX_Gamemode_Selection_Update_4D48C0(DWORD input)
 {
-    NOT_IMPLEMENTED();
-    return 0;
+    __int16 result; // ax
```
paulsapps, line 2473: you should try to remove this var if possible, looks like you could just use early/in place returns?

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -3001,33 +3074,35 @@ void MainMenuController::HandleMainMenuUpdate()
         }
 
         // Todo: change all field_10_fn_update function types to unsigned.
-        const unsigned int pageUpdateRet = (this->*(pPage->field_10_fn_update))(inputHeld);
+        const unsigned int pageUpdateReturnedCam = (this->*(pPage->field_10_fn_update))(inputHeld);
```
paulsapps, line 3077: you could probably change these into a struct, something like:

struct NextPage
{
BYTE targetCamera;
BYTE screenChangeEffect
};

?

Then you won't need the >> 16 & 0xFF and the BYTE1() on the return value

paulsapps:
ah I just noticed you said you'd do the struct stuff in another PR

Source/AliveLibAE/MainMenu.cpp
pryon: yup that will be in the next PR, along with the function and lambda return values

PR #1003, opened 2020-12-26
[AE] MainMenu refactoring, part II

pryon:
Still a lot of renaming and organizing to be done, but stopping here for now to keep the PR at a reasonable size.

### DIFF
--- a/Source/AliveLibAE/MainMenu.cpp
+++ b/Source/AliveLibAE/MainMenu.cpp
@@ -3088,15 +3088,15 @@ void MainMenuController::HandleMainMenuUpdate()
         }
 
         field_218_target_page_index = static_cast<short>(GetPageIndexFromCam_4D05A0(pageUpdateReturnedCam.page_update_camera));
-        field_21A_target_cam = pageUpdateReturnedCam.button_idx_to_highlight;
+        field_21A_target_button_index = pageUpdateReturnedCam.button_idx_to_highlight;
         isScreenTransitionNecessary = BYTE1(pageUpdateReturnedCam.page_update_camera);
     }
     else
     {
         field_1F8_page_timeout = 0;
         field_218_target_page_index = static_cast<short>(GetPageIndexFromCam_4D05A0(pPage->field_8_next_idx));
-        field_21A_target_cam = pPage->field_C_target_camera;
-        isScreenTransitionNecessary = pPage->field_A_transistion_effect;
+        field_21A_target_button_index = pPage->field_C_initial_button_idx_to_highlight;
+        isScreenTransitionNecessary = pPage->field_A_transition_effect;
     }
 
     field_21C_bDoScreenTransistionEffect = static_cast<short>(isScreenTransitionNecessary);
@@ -3220,7 +3220,7 @@ signed int MainMenuController::ChangeScreenAndIntroLogic_4CF640()
         return 1;
 
     case 2:
-        if (sMainMenuPages_561960[field_214_page_index].field_A_transistion_effect == 7)
+        if (sMainMenuPages_561960[field_214_page_index].field_A_transition_effect == 7)
         {
             char buffer[256] = {};
 
@@ -3409,8 +3409,8 @@ signed int MainMenuController::ChangeScreenAndIntroLogic_4CF640()
 
         if (sMainMenuPages_561960[field_218_target_page_index].field_18_buttons != nullptr)
         {
-            field_1FC_button_index = field_21A_target_cam;
-            if (field_21A_target_cam == -1)
+            field_1FC_button_index = field_21A_target_button_index;
+            if (field_1FC_button_index == -1)
             {
                 MainMenuButton* pButtonsIter = sMainMenuPages_561960[field_218_target_page_index].field_18_buttons;
                 short useButtonIdx = 0;
--- a/Source/AliveLibAE/MainMenu.cpp
+++ b/Source/AliveLibAE/MainMenu.cpp
@@ -2481,12 +2481,11 @@ signed int MainMenuController::PSX_Gamemode_Selection_Update_4D48C0(DWORD input)
         {
             return MainMenuCams::eGameIsLoading_ShaddapCam;
         }
-        else
-        {
-            // apparently cam 11 was originally for the cooperative mode
-            // description screen on the PSX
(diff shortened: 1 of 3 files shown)


Source/AliveLibAE/MainMenu.cpp
```diff
@@ -1569,39 +1622,39 @@ void MainMenuController::AbeMotions_Render_4D25E0(int ** ot)
...
-        return MainMenuCams::eNoChange;
+        return { MainMenuCams::eNoChange };
     }
 
     switch (field_1FC_button_index)
     {
     case 0:
-        return 0xFFFF000E;
+        return { MainMenuCams::eAbeGamespeakCam, NO_SELECTABLE_BUTTONS };
         break;
```
paulsapps, line 1643: not sure what was going on here, the breaks are pointless because of the unconditional returns ,might as well nuke them

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -1686,22 +1739,22 @@ signed int MainMenuController::Page_Front_Update_4D0720(DWORD input)
...
         case 1:
             // Quit
             Set_Anim_4D05E0(AnimIds::eAbe_GoodBye);
-            return MainMenuCams::eDummyBlankCam;
+            return { MainMenuCams::eDummyBlankCam };
         case 2:
             // Load
             field_230_target_entry_index = 0;
-            return 0xFFFF0004;
+            return { MainMenuCams::eLoadGameMenuCam, NO_SELECTABLE_BUTTONS };
```
paulsapps, line 1750: NO_SELECTABLE_BUTTONS  should be 4 here?

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -1671,7 +1724,7 @@ signed int MainMenuController::Page_Front_Update_4D0720(DWORD input)
     {
         bLongerTimeoutToNextDemo_5C1B9A = 1;
         field_1FC_button_index = 0;
-        return 0xFFFF0016;
+        return { MainMenuCams::eDemoIsLoading_ShaddapCam, NO_SELECTABLE_BUTTONS };
```
paulsapps, line 1727: NO_SELECTABLE_BUTTONS  should be 0x16 here?

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -1943,13 +1996,13 @@ EXPORT signed int MainMenuController::BackStory_Or_NewGame_Update_4D1C60(DWORD i
...
             field_1FC_button_index = 1; // Select start game
-            return MainMenuCams::eNoChange;
+            return { MainMenuCams::eNoChange };
         }
         else if (field_1FC_button_index == 1) // Start game
         {
             sCurrentControllerIndex_5C1BBE = 0;
             sGameStartedFrame_5C1B88 = sGnFrame_5C1B84;
-            return 0xFFFF000D;
+            return { MainMenuCams::eGameIsLoading_ShaddapCam, NO_SELECTABLE_BUTTONS };
```
paulsapps, line 2005: NO_SELECTABLE_BUTTONS  should be 0xD here?

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -2249,21 +2302,21 @@ signed int MainMenuController::DemoSelect_Update_4D0E10(DWORD input)
 
     if (input_or_field_204 & InputCommands::eBack)
     {
-        return 0x10003; //Esc Pressed
+        return { MainMenuCams::eOptionsCam, 1 }; //Esc Pressed
```
paulsapps, line 2305: 1 should be 3 ?

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -2249,21 +2302,21 @@ signed int MainMenuController::DemoSelect_Update_4D0E10(DWORD input)
...
     }
 
     if (input_or_field_204 & InputCommands::eUnPause_OrConfirm)
     {
         // selected a demo for playing
         gIsDemoStartedManually_5C1B9C = TRUE;
         sDemoIdChosenFromDemoMenu_5C1B9E = field_230_target_entry_index;
-        return 0xFFFF0016; //Enter Pressed
+        return { MainMenuCams::eDemoIsLoading_ShaddapCam, NO_SELECTABLE_BUTTONS }; //Enter Pressed
```
paulsapps, line 2313: NO_SELECTABLE_BUTTONS  should be 0x16?

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -3066,33 +3114,25 @@ void MainMenuController::HandleMainMenuUpdate()
...
-        field_21A_target_cam = static_cast<short>(v19);
-        if (v19 == 0xFF)
-        {
-            field_21A_target_cam = -1;
-        }
-
-        isScreenTransitionNecessary = BYTE1(pageUpdateReturnedCam);
+        field_218_target_page_index = static_cast<short>(GetPageIndexFromCam_4D05A0(pageUpdateReturnedCam.page_update_camera));
+        field_21A_target_button_index = pageUpdateReturnedCam.button_idx_to_highlight;
+        isScreenTransitionNecessary = BYTE1(pageUpdateReturnedCam.page_update_camera);
```
paulsapps, line 3128: why is BYTE1 needed? can't this be split into 2 1 byte fields?

paulsapps:
clang/gcc hate those inits

```
/home/vsts/work/1/s/Source/AliveLibAE/MainMenu.cpp: In member function ‘MainMenuNextCam MainMenuController::AbeMotions_Update_4D1F50(DWORD)’:
/home/vsts/work/1/s/Source/AliveLibAE/MainMenu.cpp:2488:42: error: could not convert ‘{eNoChange}’ from ‘<brace-enclosed initializer list>’ to ‘MainMenuNextCam’
         return { MainMenuCams::eNoChange };
```

you might need to always pass the 2nd value or something

paulsapps:
Also most of my comments about NO_SELECTABLE_BUTTONS  are probably wrong because the byte order is swapped, ignore if so

Source/AliveLibAE/MainMenu.cpp
pryon: Good point, though reading the value of `page_update_camera` isn't even needed, since the HIBYTE/BYTE1 of it will always be 0 (as MainMenuCams values only go up to 31).
As a result, `isScreenTransitionNecessary` can always be set to 0.

Issue #1007, opened 2020-12-30
AE OG Bug - Cutscenes cannot be skipped with controller input


pryon:
**Description of issue:** 
Cutscenes, FMVs and any kind of movies in the game that otherwise can be skipped with <kbd>Enter</kbd> or <kbd>Esc</kbd> cannot be skipped with any controller input when the Gamepad is selected as the input device.

In the PSX version of the game, the movies can be skipped with both the Start and the X button.

**Steps to repro:** 
1. Go to Options -> Controls
2. Select Gamepad as the input device
3. Start a new game
4. Try skipping the intro movies by pressing any button on your controller

PR #1008, opened 2020-12-30
OG bugfix #1007

pryon:


Source/AliveLibAE/Movie.cpp
```diff
@@ -311,7 +314,8 @@ EXPORT char CC DDV_Play_Impl_4932E0(const char* pMovieName)
         return 0;
     }
 
-    while (Input_IsVKPressed_4EDD40(VK_ESCAPE) || Input_IsVKPressed_4EDD40(VK_RETURN))
+    while (Input_IsVKPressed_4EDD40(VK_ESCAPE) || Input_IsVKPressed_4EDD40(VK_RETURN) ||
+        Input_Read_Pad_4FA9C0(sCurrentControllerIndex_5C1BBE) & MOVIE_SKIPPER_GAMEPAD_INPUTS)
```
paulsapps, line 318: should only be called if game pad is active? does it still make sense to check both keyboard + pad? 🤔

Issue #1083, opened 2021-01-30
Missing license


IacobusKopiirefuto:
I have found your project in this list <https://luxtorpeda-dev.github.io/packages.html> of game engines, where license for your project is listed as "Unknown". As you can see in the list, there are many reverse engineering projects of similar nature, yet they can share code under clear libre/open source license. 

Could you pick a license or at least at give some information about the copyright status in your README.md file?

paulsapps:
```
When you make a creative work (which includes code), the work is under exclusive copyright by default. Unless you include a license that specifies otherwise, nobody else can copy, distribute, or modify your work without being at risk of take-downs, shake-downs, or litigation. Once the work has other contributors (each a copyright holder), “nobody” starts including you.
```

This probably is accurate at this point in time until the code base has been refactored/changed enough. If you copied this code I can't tell you that OWI wouldn't issue a take down notice against you due to it being derived from reverse engineering. Once the code has been refactored and the engines merged and it no longer has roots in the original code I would feel more confident giving it a license at that point. Once that point is reached I would probably use the MIT license, however this would need a discussion with the team that currently works on this project. The usual "I am not a lawyer" disclaimers apply :).

IacobusKopiirefuto:
@paulsapps:
> Once the code has been refactored and the engines merged and it no longer has roots in the original code I would feel more confident giving it a license at that point.  

Sounds reasonable. Thanks for a quick reply. You might add this information to `README.md` so it's easy to find, but that's of course on you.

And remember, when you give a clear license to this project, you will most likely need confirmation from all contributors as they currently should be the copyright holders of the code that they committed. For example OpenBSD, had [problems](https://en.wikipedia.org/wiki/OpenBSD#Licensing) with this and had to remove and rewrite some of their code.

You might want to add notice to the Contributors part of README, such as
> By contributing to this repository, you agree to license your work under any of the [OSI-approved](https://opensource.org/licenses), which the core development team might decide to apply to this work in the future.

Unfortunately, I do not now how much legally biding something like that would be, but it might at least prevent future disagreements.

paulsapps:
Makes sense, PR it in if you want.

IacobusKopiirefuto:
Will do.

PR #1084, opened 2021-01-30
Adding information about License

IacobusKopiirefuto:
Result of discussion in Issue #1083 . Temporary solution to the licensing issue until a clear licensing is given.

GPL compatible licenses (license of Linux kernel, GIMP, LibreOffice or SuperTuxKart), are licenses that can be incorporated into code licensed under GPL. This includes public domain dedication such as  CC0, The Unlicense; permissive licenses such as [MIT license](https://opensource.org/licenses/MIT) (called [Expat License](https://directory.fsf.org/wiki/License:Expat) in GPL-compatible list), various BSD licenses, Apache License Version 2.0; the copyleft licenses such as LGPL, GPL. In short, almost any license you might want.

In the original issue I have used wording ["OSI-approved licenses"](https://opensource.org/licenses/category) instead of "GPL-Compatible license". However, the OSI-approved licenses also include licenses that have been superseded by newer versions, licenses that have been voluntarily retired and other, so I find GPL-compatible list more useful. Futhermore, it also includes CC0 unlike the [OSI-approved licenses](https://opensource.org/faq#cc-zero) list.

You might also want to specify, who is part of the "core development team" and get written agreement from all the past contributors that they agree with these  temporary licensing terms. For example such as separate Issue, to which they all simply reply "I agree that my past contribution will be licensed under these terms.".

I am **not** lawyer and this is **not** legal advice. If you have any questions regarding the best license for your code or any other legal issues relating to it, please do further research or consult with a professional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,12 @@ https://discord.gg/khs6KKS
 ## Contributing
 If you'd like to contribute to our reverse engineering efforts, make sure to follow our reverse engineering guidelines:
 
+### License
+
+So far the code is **not** shared under any license. Once the code has been refactored and the engines merged and it no longer has roots in the original code we would feel more confident giving it a license at that point.
+
+Until then, by contributing to this repository, you agree to license your work under any of the [GPL-Compatible](https://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses) licenses, which the core development team might decide to apply to this work in the future.
+
 ### IDA Guidelines
 - Use IDA 7
 - Use our git hosted IDB.


paulsapps:
-  @d3nwah 
- @mouzedrift 
- @mlgthatsme 
- @pryon 
- @UltraStars3000 
- @Kari1993 
- @GproKaru 
- @FrozenFish24 
- @JayPeet 
- @JonathanPicques 
- @AvengingGibbons

Nemin32:
Hi,
We edited the README file to reflect the project's approach to an eventual licensing. However, without an actual legal expert's opinion, we're currently unwilling to make further statements about the license as frankly no one wants to get into trouble for saying something that turns out to be dumb.

If you've anything else to recommend please feel free to reply, otherwise just close the request.

IacobusKopiirefuto:
That's all for now. I'm glad that I could help.

PR #1095, opened 2021-02-01
don't boot to the save screen in AE by default in appveyor builds

paulsapps:


paulsapps:
(hold shift in windows to still boot to the save load screen)

PR #1103, opened 2021-02-04
build release (avoid issue with masher videos running too slowly)

paulsapps:


paulsapps:
closing this - I wanted to compile certain files as release but CMake makes this very difficult. Instead of building everything for release it is probably better to fix this issue by pushing ahead with the assets conversion and using ffmpeg or the likes to convert ddv to mp4/h264.

Issue #1010, opened 2020-12-31
AO Rev Bug - Controls menu rebinds wrong actions


rozniak:
Found in build: 54f7418412105462e363123380ed9719ee4ae344
Last working build number: N/A

Description of issue:
Rebinding controls using the in game menu rebinds the wrong things:
- Run rebinds right
- Sneak rebinds down
- Jump rebinds nothing
- Action/Throw are the wrong way around
- Crouch rebinds sneak

Editing `abe2.ini` manually works, it's the in game menu that is broken.

Here's my usual control scheme:
```
run = Q
sneak = A
jump = W
action = S
throw = O
fart = E
```

Here's what I ended up with when I tried this binding in game:
```
run = W
sneak = E
jump =
action = O
throw = S
fart =
```

Steps to repro:
Rebind controls using the in game menu - notice that menu navigation using right and down keys use rebound keys and in game controls are very mixed up.

paulsapps:
@pryon likes fixing these sorts of bugs 😏

pryon:
Yeah the menus in AO are kinda borked, a big menu overhaul is planned 😉

pryon:
@paulsapps this can be closed now

PR #1115, opened 2021-02-07
Massive AO input handling rework and *_IMPLEMENTED macro nuking

pryon:
Also includes:
 - tweaks in remapping menu to behave like AE
 - a fix for #1010 
 - applied input enums
 - renames


Source/AliveLibAO/CheatController.cpp
```diff
@@ -63,12 +63,12 @@ const InputCommands sCheatKeyArray_FartGas_4C50B8[] =
...
 const InputCommands sCheatKeyArray_VoiceLocks_4C50C8[] =
 {
-    eFartOrRoll,
+    eCrouchOrRoll,
     eDown,
     eThrowItem,
     eDoAction,
     eDown,
-    eFartOrRoll,
+    eCrouchOrRoll,
```
paulsapps, line 71: isn't this here twice?

pryon: this is a cheat button sequence, I think multiple entries are fine here (eDown is also here twice)

PR #1168, opened 2021-03-17
Add and use scale enum for AE

mouzedrift:


Source/AliveLibAE/GlukkonSwitch.cpp
```diff
@@ -29,12 +29,13 @@ GlukkonSwitch* GlukkonSwitch::ctor_444E60(Path_GlukkonSwitch* pTlv, int tlvInfo)
     field_F4_tlvInfo = tlvInfo;
     field_20_animation.field_C_render_layer = Layer::eLayer_22;
 
-    if (pTlv->field_10_scale == 1)
+    // TODO: new scale enum because the values are the wrong way around
+    if (pTlv->field_10_scale == Scale_short::eHalf_1)
```
paulsapps, line 33: I think you should leave these ones as ints or change them now because this is crazy confusing and contradicts the logic

paulsapps: .. unless this is a rev bug ?

mouzedrift: not a rev bug

paulsapps: pTlv->field_10_scale should be changed to another enum like Inverted_Scale_short or something

PR #1201, opened 2021-03-24
always --gtest_catch_exceptions=0

paulsapps:


Source/Tools/alive_api/alive_api_test.cpp
```diff
@@ -58,6 +58,7 @@ TEST(alive_api, ExportPathBinaryToJsonAE)
 {
     auto ret = AliveAPI::ExportPathBinaryToJson("OutputAE.json", AEPath(kAETestLvl), 14);
     ASSERT_EQ(ret.mResult, AliveAPI::Error::None);
+    abort();
```
mouzedrift, line 61: is this intentional? :tape:

paulsapps: ⚰️

Issue #1018, opened 2021-01-01
AE Rev Bug - chanting while scrab is turning turns them peaceful


kamil-sita:
Found in build:  3455 from AppVeyor, which is probably 54f7418
Last working build number: ?

Description of issue:
Chanting with turning scrab on screen turns them peaceful.

Steps to repro:
1. (?) Get to SVP02C06
2. Chant while scrab is turning - somewhat hard to time it correctly
3. Scrab is now peaceful

video:

https://user-images.githubusercontent.com/24254781/103461568-925faa00-4d1f-11eb-92f1-c55414eaf544.mp4

save: 
[SVP02C06-buggy.zip](https://github.com/AliveTeam/alive_reversing/files/5759298/SVP02C06-buggy.zip)

kamil-sita:
Added video

PR #1203, opened 2021-03-24
#1018 fix. Scrab chase logic less janky

d3nwah:
https://github.com/AliveTeam/alive_reversing/issues/1018

### DIFF
--- a/Source/AliveLibAE/Scrab.cpp
+++ b/Source/AliveLibAE/Scrab.cpp
@@ -1253,7 +1253,7 @@ __int16 Scrab::AI_ChasingEnemy_1_4A6470()
     }
 
     LiftPoint* pLiftPoint = static_cast<LiftPoint*>(sObjectIds_5C1B70.Find_449CF0(field_110_id));
-    if (pLiftPoint->field_4_typeId != Types::eLiftPoint_78)
+    if (pLiftPoint && pLiftPoint->field_4_typeId != Types::eLiftPoint_78)
     {
         pLiftPoint = nullptr; //OG bug fix: Before it could use the pointer as a LiftPoint even if it, in fact, wasn't one
     }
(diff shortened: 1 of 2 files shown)


Source/AliveLibAE/Scrab.cpp
```diff
@@ -1634,52 +1614,50 @@ __int16 Scrab::AI_ChasingEnemy_State_Running_2(BaseAliveGameObject* pObj)
...
-            {
-                field_108_next_motion = eScrabMotions::M_Turn_3_4A91A0;
-            }
-            else
+            if (field_106_current_motion != eScrabMotions::M_JumpToFall_8_4A9220)
             {
-                field_106_current_motion = eScrabMotions::M_Turn_3_4A91A0;
-                field_108_next_motion = -1;
-                MapFollowMe_408D10(TRUE);
+                //A patch workaround
```
paulsapps, line 1638: what does this mean?

Source/AliveLibAE/Scrab.cpp
```diff
@@ -1246,15 +1246,18 @@ __int16 Scrab::AI_ChasingEnemy_1_4A6470()
...
         return AI_ChasingEnemy::eState1_Panic_4;
     }
 
-    LiftPoint* pLiftPoint = static_cast<LiftPoint*>(pObj);
-    if (pLiftPoint && pLiftPoint->field_4_typeId == Types::eLiftPoint_78 && 
-        !pLiftPoint->vOnAnyFloor_461920() && field_11C_sub_state != AI_ChasingEnemy::eState1_Panic_4)
+    LiftPoint* pLiftPoint = static_cast<LiftPoint*>(sObjectIds_5C1B70.Find_449CF0(field_110_id));
+    if (!pLiftPoint || pLiftPoint->field_4_typeId != Types::eLiftPoint_78)
+    {
+        pLiftPoint = nullptr;
```
paulsapps, line 1258: comment that this is an OG fix for pLiftPoint being used later even if it isn't a lift point

Source/AliveLibAE/Scrab.cpp
d3nwah: It's just that this scenario should probably never happen in the first place, the condition is like a plaster on it.

d3nwah: I forgot to finish that comment :tapeDan:

paulsapps: maybe reword it cause I had no idea what it meant ⚰️

PR #1213, opened 2021-03-28
Enumify some AO TLVs

FrozenFish24:


Source/AliveLibAO/RollingBallStopper.cpp
```diff
@@ -22,15 +22,15 @@ EXPORT RollingBallStopper* RollingBallStopper::ctor_43BCE0(Path_RollingBallStopp
     
     field_114_switch_on_id = pTlv->field_18_id_on;
 
-    if (pTlv->field_1A_scale_background)
+    if (pTlv->field_1A_scale_background == Scale_short::eHalf_1)
```
mouzedrift, line 25: ``field_1A_scale_background`` can be renamed to ``field_1A_scale``

FrozenFish24: 👍

PR #1214, opened 2021-03-28
fix double knockback sound when being hit by a moving bomb and fix sAlertedMudCount_5C3010 being capped

mouzedrift:


Source/AliveLibAE/Mudokon.cpp
```diff
@@ -6983,12 +6983,13 @@ void Mudokon::M_TurnWheelEnd_59_474D30()
 
 __int16 Mudokon::StableDelay_477570()
 {
+    int maxIdx = sAlertedMudCount_5C3010;
     if (sAlertedMudCount_5C3010 >= ALIVE_COUNTOF(kDelayTable_55CF7C))
```
paulsapps, line 6987: you should probably use maxIdx here too

paulsapps: actually why is this used? why isn't it just if (sDelayIdx_5C3014 >= ALIVE_COUNTOF(kDelayTable_55CF7C)) { sDelayIdx_5C3014 = 0; } ?

PR #1238, opened 2021-04-21
Fix some readability warnings

SuperV1234:
- Swallow semicolons in macros to avoid "extra semicolon" warnings
- Wrap pieces of long logical expressions in parentheses to avoid "logical-op" warnings, making precedence obvious
- Use `[[fallthrough]]` attribute

None of these changes should have any effect on the behavior of the project whatsoever.

### DIFF
--- a/Source/AliveLibAE/Input.cpp
+++ b/Source/AliveLibAE/Input.cpp
@@ -1992,7 +1992,7 @@ void InputObject::Update_45F040()
         {
             const DWORD command = (*field_30_pDemoRes)[field_34_demo_command_index++];
             field_3C_command = command >> 16;
-            field_40_command_duration = (sGnFrame_5C1B84 + command) & 0xFFFF;
+            field_40_command_duration = sGnFrame_5C1B84 + (command & 0xFFFF);
 
             // End demo/quit command
             if (command & 0x8000)
--- a/Source/AliveLibAE/Input.cpp
+++ b/Source/AliveLibAE/Input.cpp
@@ -1992,7 +1992,7 @@ void InputObject::Update_45F040()
         {
             const DWORD command = (*field_30_pDemoRes)[field_34_demo_command_index++];
             field_3C_command = command >> 16;
-            field_40_command_duration = sGnFrame_5C1B84 + command & 0xFFFF;
+            field_40_command_duration = (sGnFrame_5C1B84 + command) & 0xFFFF;
 
             // End demo/quit command
             if (command & 0x8000)
--- a/Source/AliveLibAO/PauseMenu.cpp
+++ b/Source/AliveLibAO/PauseMenu.cpp
@@ -147,7 +147,7 @@ ALIVE_VAR(1, 0x5080C6, saveName, saveNameBuffer_5080C6, {});
 
 const char *gLevelNames_4CE1D4[20] =
 {
-    "¸",
+    "ï¿½",
     "RuptureFarms",
     "Monsaic Lines",
     "Paramonia",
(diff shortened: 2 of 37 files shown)


Source/AliveLibAE/Input.cpp
```diff
@@ -1992,7 +1992,7 @@ void InputObject::Update_45F040()
         {
             const DWORD command = (*field_30_pDemoRes)[field_34_demo_command_index++];
             field_3C_command = command >> 16;
-            field_40_command_duration = sGnFrame_5C1B84 + command & 0xFFFF;
+            field_40_command_duration = (sGnFrame_5C1B84 + command) & 0xFFFF;
```
paulsapps, line 1995: oh, this one actually isn't right, its supposed to be sGnFrame_5C1B84 + (command & 0xFFFF); as you can see the other 16bits are being used as the command above, and these are the time/duration.

Source/AliveLibAO/PauseMenu.cpp
```diff
@@ -147,7 +147,7 @@ ALIVE_VAR(1, 0x5080C6, saveName, saveNameBuffer_5080C6, {});
 
 const char *gLevelNames_4CE1D4[20] =
 {
-    "�",
+    "�",
```
paulsapps, line 150: github is displaying this in a strange way, wasn't it just a period previously?

Source/AliveLibCommon/FunctionFwd.hpp
```diff
@@ -7,15 +7,15 @@ bool RunningAsInjectedDll();
 #define CC __cdecl
 #define CCSTD __stdcall
 #if _WIN64
-#define ALIVE_ASSERT_SIZEOF(structureName, expectedSize)
+#define ALIVE_ASSERT_SIZEOF(structureName, expectedSize) static_assert(true)
```
paulsapps, line 10: is there an more idiomatic way of having a "blank" macro than this?

Source/AliveLibAE/Input.cpp
SuperV1234: I can fix it, but it's not the same behavior as prior to my PR . `+` has higher precedence than `&`.

Source/AliveLibAO/PauseMenu.cpp
SuperV1234: Not sure what happened here, will fix.

Source/AliveLibCommon/FunctionFwd.hpp
SuperV1234: Not really, especially when it's not a function-scope macro. The alternative is something like `struct DUMMY ## __LINE__ { }` which I really do not like.

Leaving the macro completely black does not swallow the semicolon and makes code that forgets the semicolon compile in `_WIN64` mode.

Source/AliveLibAO/PauseMenu.cpp
SuperV1234: This was not a period, it was `¸`. Do we want to keep it like that?
https://unichar.app/web/char/?category=Punctuation&char=%C2%B8

Source/AliveLibAE/Input.cpp
paulsapps: yeah its a bug - I do wonder how demo play back was working now though 😂

Source/AliveLibAO/PauseMenu.cpp
paulsapps: It looks like it should just be a blank string "" looks like a reversing mistake/bug

PR #1239, opened 2021-04-21
Enable CMake PCH

SuperV1234:
```bash
# without PCH, clang++ on MSYS2
make clean; time make -j8 AliveExeAO AliveExeAE AliveExe
> 2:01.45 total
```

```bash
# with PCH, clang++ on MSYS2
make clean; time make -j8 AliveExeAO AliveExeAE AliveExe
> 37.270 total
```

That's a 84s improvement for me. :)

PR #1240, opened 2021-04-21
Improve PCH performance

SuperV1234:
**Merge after #1239**

```bash
# with PCH, clang++ on MSYS2, prior to this PR
make clean; time make -j8 AliveExeAO AliveExeAE AliveExe
> 37.270 total
```

```bash
# with PCH, clang++ on MSYS2, after this PR
make clean; time make -j8 AliveExeAO AliveExeAE AliveExe
> 25.806 total
```

### DIFF
--- a/Source/AliveLibAE/DDraw.cpp
+++ b/Source/AliveLibAE/DDraw.cpp
@@ -306,7 +306,7 @@ EXPORT signed int CC DD_Shutdown_4F0790(int bDestroyDD)
             sDD_Pal_BBC3D8->Release();
             sDD_Pal_BBC3D8 = nullptr;
         }*/
-        
+
         if (bDestroyDD)
         {
             sDDraw_BBC3D4->RestoreDisplayMode();
@@ -351,7 +351,7 @@ EXPORT LPDIRECTDRAWSURFACE CC DD_Create_Surface_4F0CB0(int width, int height, in
         colourKey.dwColorSpaceHighValue = sDDColourKey_BBC3C0;
         pSurface->SetColorKey(DDCKEY_SRCBLT, &colourKey); // 8
     }
-    
+
     return pSurface;
 }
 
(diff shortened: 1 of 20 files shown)


Source/AliveLibAE/DDraw.cpp
```diff
@@ -625,7 +625,7 @@ EXPORT signed int CC DD_Enable_4F0380(HWND /*hwnd*/, int width, int height, int
         ::ReleaseDC(nullptr, dc);
 
         const LONG oldWinStyle = ::GetWindowLongA(sDD_hWnd_BBC3B0, GWL_STYLE);
-#if BEHAVIOUR_CHANGE_FORCE_WINDOW_MODE
+#ifdef BEHAVIOUR_CHANGE_FORCE_WINDOW_MODE
```
paulsapps, line 628: I'm not sure this is correct - BEHAVIOUR_CHANGE_FORCE_WINDOW_MODE is always defined by cmake as 

#define BEHAVIOUR_CHANGE_FORCE_WINDOW_MODE 0
or
#define BEHAVIOUR_CHANGE_FORCE_WINDOW_MODE 1

so now it appears that this block will always be compiled in no matter the macro value

Source/AliveLibAE/LvlArchive.cpp
```diff
@@ -94,7 +94,7 @@ int LvlArchive::Open_Archive_432E80(const char* fileName)
 
     // Open the LVL file
 
-#if BEHAVIOUR_CHANGE_SUB_DATA_FOLDERS
+#ifdef BEHAVIOUR_CHANGE_SUB_DATA_FOLDERS
```
paulsapps, line 97: ditto for all these I guess

Source/AliveLibAE/DDraw.cpp
SuperV1234: Yes, my bad. I was playing around with these to figure out the invalid PCH issue and forgot to turn them back.

PR #1242, opened 2021-04-21
More PCH improvements (MSVC) 

SuperV1234:


Source/AliveLibAE/CMakeLists.txt
```diff
@@ -415,11 +413,7 @@ target_compile_features(AliveLibAE
     PRIVATE cxx_variadic_templates)
 
 if (MINGW)
-    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-        set(WIN32_LIBS "imm32")
-    else()
-        set(WIN32_LIBS "")
-    endif()
+    set(WIN32_LIBS "imm32")
```
SuperV1234, line 416: Turns out this is needed for both gcc and clang when doing PCH builds, dunno but it works.

paulsapps: strange how the hosted CI is working without too

SuperV1234: does the hosted CI even build with MinGW?

paulsapps: nope its MSVC, clang and gcc (so I would have expected the clang build to fail?)

SuperV1234: clang/gcc need imm32 only for MinGW

PR #1246, opened 2021-04-22
Added chanting with any shoulder button combination

deanff:


Source/AliveLibAE/Input.cpp
```diff
@@ -1379,13 +1382,34 @@ EXPORT int Input_Convert_KeyboardGamePadInput_To_Internal_Format_492150()
...
 
-                    if (input_command_c_pressed)
+                    if (shoulderButtonsPressedCount > 1) // Added functionality for PC - chant with any shoulder button combo
+                    {
+                        pressed_keyboard_keys |= InputCommands::Enum::eChant;
+                    }
+                    else if (input_command_c_pressed)
                     {
-                        if (input_command_delete_pressed)
+                        if (input_command_delete_pressed) // Old way of chanting on PC - no longer called - kept for preservation
```
paulsapps, line 1412: if there is any point of this if its dead code?

Source/AliveLibAE/Input.cpp
```diff
@@ -1362,6 +1362,8 @@ EXPORT int Input_Convert_KeyboardGamePadInput_To_Internal_Format_492150()
                 {
                     buttons = pButtons;
 
+                    unsigned int shoulderButtonsPressedCount = 0; // Added for chanting with any shoulder button combo
```
paulsapps, line 1365: I would add a comment marking this as an OG change so it can easily be looked/up spotted when fixing any reversing bugs

Source/AliveLibAE/Input.cpp
deanff: Wasn't sure if anyone needed the old functionality for testing side-by-side, or preservation. A bool could be added to switch between the two. Alternatively, it could just be removed.

Source/AliveLibAE/Input.cpp
deanff: Noted. Should I say "OG change" for each comment, just here, or maybe on the if(shoulderButtonsPressedCount > 1) line, where the change mostly occurs?

paulsapps: if you're gonna keep both I'd just have an IsChanting() and IsChantingAnyShoulderButton() method or something and put them in that, then the call it only needs the comment and not every line

Source/AliveLibAE/Input.cpp
```diff
@@ -1264,6 +1264,24 @@ ALIVE_VAR(1, 0x5c9f74, DWORD, sPrevious_down_keyboard_keys_5C9F74, 0);
 ALIVE_VAR(1, 0x5c9f78, DWORD, dword_5C9F78, 0);
 ALIVE_VAR(1, 0x5c9794, int, sKeyboardBindings_5C9794, 0);
 
+bool IsChantingAnyShoulderButton(int shoulderButtonsPressedCount)
+{
+    if (shoulderButtonsPressedCount > 1)
```
paulsapps, line 1269: can just be return shoulderButtonsPressedCount > 1; ? this function can be static too

Source/AliveLibAE/Input.cpp
```diff
@@ -1264,6 +1264,24 @@ ALIVE_VAR(1, 0x5c9f74, DWORD, sPrevious_down_keyboard_keys_5C9F74, 0);
...
+    {
+        return true;
+    }
+    return false;
+}
+
+bool IsChanting(char input_command_c_pressed, char input_command_delete_pressed)
```
paulsapps, line 1276: cane be static

Source/AliveLibAE/Input.cpp
```diff
@@ -1264,6 +1264,24 @@ ALIVE_VAR(1, 0x5c9f74, DWORD, sPrevious_down_keyboard_keys_5C9F74, 0);
...
+{
+    if (input_command_c_pressed && input_command_delete_pressed)
```
paulsapps, line 1278: just return input_command_c_pressed && input_command_delete_pressed; ?

Source/AliveLibAE/Input.cpp
```diff
@@ -1379,38 +1400,56 @@ EXPORT int Input_Convert_KeyboardGamePadInput_To_Internal_Format_492150()
...
-                                pressed_keyboard_keys |= InputCommands::Enum::eGameSpeak1;
+                                ++shoulderButtonsPressedCount;
                             }
-                            buttons = pButtons & ~(InputCommands::Enum::eRight | InputCommands::Enum::eLeft | InputCommands::Enum::eDown | InputCommands::Enum::eUp);
-                            pButtons &= ~(InputCommands::Enum::eRight | InputCommands::Enum::eLeft | InputCommands::Enum::eDown | InputCommands::Enum::eUp);
                         }
                     }
+
+                    // OG Change - chant with any shoulder button combo
+                    bool isChanting = IsChantingAnyShoulderButton(shoulderButtonsPressedCount);
```
paulsapps, line 1425: can be const

Source/AliveLibAE/Input.cpp
```diff
@@ -1379,38 +1400,56 @@ EXPORT int Input_Convert_KeyboardGamePadInput_To_Internal_Format_492150()
...
-                            if (pButtons & InputCommands::Enum::eUp)
-                            {
-                                pressed_keyboard_keys |= InputCommands::Enum::eGameSpeak2;
-                            }
-                            if (pButtons & InputCommands::Enum::eDown)
-                            {
-                                pressed_keyboard_keys |= InputCommands::Enum::eGameSpeak3;
-                            }
-                            if (pButtons & InputCommands::Enum::eLeft)
+                            if ((1 << i) & pButtons)
```
paulsapps, line 1409: can we not explicitly check the required button bits instead of doing it this way? this way is harder to understand what is happening

deanff: Just to clarify - are you asking me to modify the way the game checks for eSpeak1 and eSpeak2?
Note: I haven't modified this - I've just added a check for eRun and eSneak in the same loop.

paulsapps: I'm talking about ` if ((1 << i) & pButtons)` which seems to be a new addition?

deanff: That was used for eSpeak1 and eSpeak2 as well. I've just followed its implementation.

Source/AliveLibAE/Input.cpp
```diff
@@ -1379,37 +1392,48 @@ EXPORT int Input_Convert_KeyboardGamePadInput_To_Internal_Format_492150()
...
                             }
                         }
+
+                        if (sGamePadBindings_5C98E0[i] & InputCommands::Enum::eRun)
+                            if ((1 << i) & pButtons)
+                                ++shoulderButtonsPressedCount;
+
+                        if (sGamePadBindings_5C98E0[i] & InputCommands::Enum::eSneak)
+                            if ((1 << i) & pButtons)
+                                ++shoulderButtonsPressedCount;
```
paulsapps, line 1405: need braces for single line statements

Source/AliveLibAE/Input.cpp
```diff
@@ -1379,37 +1392,48 @@ EXPORT int Input_Convert_KeyboardGamePadInput_To_Internal_Format_492150()
...
                     }
 
-                    if (input_command_c_pressed)
+                    // OG Change - chant with any shoulder button combo
+                    const bool isChanting = IsChantingAnyShoulderButton(shoulderButtonsPressedCount);
+                    // Original Method:
+                    IsChanting(input_command_c_pressed, input_command_delete_pressed);
```
paulsapps, line 1411: I'm confused because now it calls both?  prob need something in options.cmake for this?

PR #1247, opened 2021-04-22
Renamed AliveExe to relive, fixed Linux scripts and RPM generation

Nemin32:


Source/AliveExe/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ add_executable(AliveExe WIN32
     resource.rc
     resource.h
     app.ico)
+  set_target_properties(AliveExe PROPERTIES OUTPUT_NAME "relive")
```
paulsapps, line 12: can't we just rename AliveExe to actually be relive instead?

Nemin32: We can

PR #1250, opened 2021-04-24
Exit save menu using controller

### LINKED ISSUE #1173: AE OG Bug - Save screen cannot be quit with controller
**Description of issue:** 
The save screen in the pause menu cannot be quit with any of the usual controller inputs, pressing <kbd>Esc</kbd> on the keyboard is necessary to get out.

Expected behavior would be to exit with Y/Triangle or Start.

Reportedly OG Oddysee lets you get out of this screen by pressing Start or Select on the controller, but I haven't been able to verify this with my virtual XInput controller, so this would need fixing on the Oddysee side as well.

**Steps to repro:** 
1. Set the controller as the active input device
2. Start a new game
3. Pause the game, select *Save*
4. Try quitting with any of the controller buttons

deanff:
Fixes #1173

### DIFF
--- a/Source/AliveLibAE/Sys.cpp
+++ b/Source/AliveLibAE/Sys.cpp
@@ -29,6 +29,15 @@ ALIVE_VAR(1, 0xBBB9EC, HINSTANCE, sInstance_BBB9EC, nullptr);
 ALIVE_VAR(1, 0xBBB9FC, int, sCmdShow_BBB9FC, 0);
 ALIVE_VAR(1, 0xBBFB04, TWindowHandleType, hWnd_BBFB04, nullptr);
 
+#if ORIGINAL_PS1_BEHAVIOR
+bool saveMenuOpen = false;
+
+EXPORT void setSaveMenuOpen(bool val)
+{
+    saveMenuOpen = val;
+}
+#endif
+
 EXPORT void CC Sys_Set_Hwnd_4F2C50(TWindowHandleType hwnd)
 {
     hWnd_BBFB04 = hwnd;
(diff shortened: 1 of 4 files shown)


Source/AliveLibAE/Sys.cpp
```diff
@@ -29,6 +29,15 @@ ALIVE_VAR(1, 0xBBB9EC, HINSTANCE, sInstance_BBB9EC, nullptr);
 ALIVE_VAR(1, 0xBBB9FC, int, sCmdShow_BBB9FC, 0);
 ALIVE_VAR(1, 0xBBFB04, TWindowHandleType, hWnd_BBFB04, nullptr);
 
+#if ORIGINAL_PS1_BEHAVIOR
+bool saveMenuOpen = false;
```
paulsapps, line 33: this can be static

Source/AliveLibAO/PauseMenu.cpp
```diff
@@ -384,6 +392,24 @@ void PauseMenu::VUpdate_44DFB0()
 
                     auto last_pressed = Input_GetLastPressedKey_44F2C0();
                     char lastPressedKeyNT[2] = { last_pressed, 0 };
+
+#if ORIGINAL_PS1_BEHAVIOR // OG Change - Exit save menu using controller
+                    if (last_pressed == VK_ESCAPE || last_pressed == VK_RETURN) // Keyboard ESC or ENTER
```
paulsapps, line 397: this whole block is duplicated earlier - should prob be in its own func

deanff: Should I make the function in AE or AO?

paulsapps: Oh yeah I forgot everything exists twice - yeah just leave it copy pasted for now I guess

PR #1260, opened 2021-04-25
Fixes Auto-Turn Bug

### LINKED ISSUE #1032: AO+AE OG Bug - Traditionally known as the "auto-turn" bug
https://user-images.githubusercontent.com/32264266/104122405-be40e800-533c-11eb-98c2-019c46e0e172.mp4

Anywhere you perform an action that requires Abe to be facing a specific direction (ie. using lifts in AO, mounting elum, hoist up or down ledges) is bugged. If Abe is not in the correct orientation, the game will correct his orientation by forcing a turn animation. The action process can be interrupted whilst Abe is turning around by running into the direction Abe is turning to face. The action will remain in queue waiting for Abe to do his next turn animation (irrelevant of direction or whethere standing or crouched) to carry out the action. This allows the player to travel where possible to a location where they wish for Abe to carry out the action, allowing for things like hoisting down through floors.

Stand on the edge of a ledge or underneath the edge of a ledge (ledge must be one where you can hoist up or down) facing away from the ledge. Press up or down to hoist, but immediately run into the direction of the ledge as Abe is turning in order to prevent the hoisting action. Then turn around and watch Abe attempt to hoist in a place other than where the ledge is.

deanff:
Fixes #1032


Source/AliveLibAE/Abe.cpp
```diff
@@ -3581,6 +3581,11 @@ void Abe::State_2_StandingTurn_451830()
 
     if (currentFrame == 4 && (sInputKey_Run_5550E8 & pressed) && (sInputKey_Right_5550D0 | sInputKey_Left_5550D4) & pressed)
     {
+
+#if ORIGINAL_GAME_FIXES
+        field_108_next_motion = 0; // OG Change - Fixes "Auto-Turn" bug
+#endif
```
paulsapps, line 3587: I think there should be an additional cmake option for these gameplay bugs for now - since the speed runners will want to keep some (certainty this one).
 Also there should be motion enum value for 0, it should be idle.

Source/AliveLibAO/Abe.cpp
```diff
@@ -4323,6 +4323,11 @@ void Abe::State_2_StandingTurn_426040()
         if (Input().IsAnyPressed(sInputKey_Run_4C65A8) &&
             Input().IsAnyPressed(sInputKey_Right_4C6590 | sInputKey_Left_4C6594))
         {
+
+#if ORIGINAL_GAME_FIXES
+            field_FE_next_state = 0; // OG Change - Fixes "Auto-Turn" bug
```
paulsapps, line 4328: ditto for enum value

Source/AliveLibAE/Abe.cpp
deanff: I created "ORIGINAL_GAME_FIXES" in cmake.options for this bugfix. Or do you mean that you'd prefer it set off by default?

paulsapps: I mean it should be possible to set it per bug, so have ORIGINAL_GAME_FIXES = all fixes and a specific options for ORIGINAL_GAME_FIX_AUTO_TURN and have it so if either is enabled then its fixed, we'd want this for all "speed run strat" fixes

PR #1261, opened 2021-04-25
`relive_api` refactoring and compilation/linking speed improvements

SuperV1234:
Links around 10x faster on a MinGW + MSYS2 + GCC build environment.


Source/Tools/relive_api/AETlvs.hpp
```diff
@@ -256,9 +258,9 @@ namespace AETlvs
 
         CTOR_AE(Path_ElectricWall, "ElectricWall", TlvTypes::ElectricWall_38)
         {
-            ADD("Scale", mTlv.field_10_scale);
-            ADD("ID", mTlv.field_12_id);
-            ADD("Start State", mTlv.field_14_start_state);
+            TLVOBJECTBASE_ADD("Scale", mTlv.field_10_scale);
```
paulsapps, line 261: does this have to be renamed? its super verbose now and it hurts the readability somewhat

SuperV1234: Doesn't have to, but I think it's an improvement, `ADD` is too short of a name for a macro definition. Could easily collide with new macros in the future or if we add a new 3rd party dependency. Also makes it clear where the macro is defined.

Source/Tools/relive_api/TlvObjectBaseMacros.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include "TlvObjectBase.hpp"
+
+// #define ADD_PROP(name, type) AddProperty(name, globalTypes.TypeName(typeid(type)), &type)
```
paulsapps, line 5: might as well delete all these commented lines if they're no longer used

Source/Tools/relive_api/AETlvs.hpp
paulsapps: I think it should be left as ADD for now - could just add a comment in each of those AO/AE headers documenting where it comes from given its used in like 90% of the lines that make up these 2 files

paulsapps:
```c
    [[nodiscard]] const std::string& ValueToString(T valueToFind) const
    {
        for (const auto [key, value] : mMapping)
        {
            if (key == valueToFind)
            {
                return value;
            }
        }

        throw ReliveAPI::UnknownEnumValueException();
    }
    ```

this fails as the return is returning a reference to an out of scope var

**Comment by paulsapps** [comment]
I removed the references from things that had warnings to get it to build and then the tests failed:

```
[ RUN      ] alive_api.ImportPathJsonToBinaryAO
E:\Data\alive\reversing\alive_reversing_paulsapps\alive_reversing\Source\Tools\relive_api\relive_api_test.cpp(104): error: Expected equality of these values:
  ogR1
    Which is: { '\0', '(' (40, 0x28), '\0', '\0', '\0', '\0', '\0', '\0', 'I' (73, 0x49), 'n' (110, 0x6E), 'd' (100, 0x64), 'x' (120, 0x78), '\0', '\0', '\0', '\0', '\xD2' (210), '\0', '\0', '\0', '\x5' (5), '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', ... }
  rewrittenR1
    Which is: { '\0', '(' (40, 0x28), '\0', '\0', '\0', '\0', '\0', '\0', 'I' (73, 0x49), 'n' (110, 0x6E), 'd' (100, 0x64), 'x' (120, 0x78), '\0', '\0', '\0', '\0', '\xD2' (210), '\0', '\0', '\0', '\x5' (5), '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', ... }
[  FAILED  ] alive_api.ImportPathJsonToBinaryAO (84 ms)
```

SuperV1234:
> ```cpp
>     [[nodiscard]] const std::string& ValueToString(T valueToFind) const
>     {
>        ...
>     }
> ```
> this fails as the return is returning a reference to an out of scope var 
> ```

ouch. This is why #1264 is important. Will fix

PR #1264, opened 2021-04-25
Enable GCC/Clang warnings and fix some of them

SuperV1234:


Source/AliveLibAE/BirdPortal.cpp
```diff
@@ -763,11 +763,15 @@ s32 CC BirdPortal::CreateFromSaveState_499C90(const u8* pBuffer)
 
     case PortalStates::ShrykullGetDoves_7:
     case PortalStates::Unused_8:
+    {
         if (savedState == PortalStates::Unused_8)
         {
             ALIVE_FATAL("PortalStates::Unused_8 never expected to be used");
         }
 
+        break;
```
paulsapps, line 772: This probably needs to be checked against the original as case 8 previously fell through, is what what OG does?

Source/AliveLibAE/NakedSlig.cpp
```diff
@@ -1305,7 +1305,7 @@ s16 NakedSlig::AI_4_GetKilled_41A880()
             {
                 Slig_GameSpeak_SFX_4C04F0(
                     SligSpeak::eHelp_10,
-                    static_cast<s16>(2 * (field_1AC_timer & 0xFFFF - sGnFrame_5C1B84)),
+                    static_cast<s16>(2 * (field_1AC_timer & (0xFFFF - sGnFrame_5C1B84))),
```
paulsapps, line 1308: this seems somewhat suspect, not even sure what this is trying to calculate

Source/AliveLibAO/Input.cpp
```diff
@@ -312,7 +312,7 @@ EXPORT void InputObject::Update_433250()
         {
             const u32 command = (*field_18_demo_res)[field_1C_demo_command_index++];
             field_24_command = command >> 16;
-            field_28_command_duration = gnFrameCount_507670 + command & 0xFFFF;
+            field_28_command_duration = (gnFrameCount_507670 + command) & 0xFFFF;
```
paulsapps, line 315: same bug we saw in AE, the &0xFFFF is supposed to be on `command` as its being split into hi and lo

Source/AliveLibAO/MainMenu.cpp
```diff
@@ -1092,7 +1092,7 @@ void Menu::FMV_Select_Update_47E8D0()
                     gPsxDisplay_504C78.PutCurrentDispEnv_40DE40();
                     pScreenManager_4FF7C8->DecompressCameraToVRam_407110(reinterpret_cast<u16**>(gMap_507BA8.field_34_camera_array[0]->field_C_ppBits));
                     pScreenManager_4FF7C8->MoveImage_406C40();
-                    pScreenManager_4FF7C8->field_36_flags = pScreenManager_4FF7C8->field_36_flags & ~1 ^ 1; // Set 1 ?
+                    pScreenManager_4FF7C8->field_36_flags = (pScreenManager_4FF7C8->field_36_flags & ~1) ^ 1; // Set 1 ?
```
paulsapps, line 1095: should prob remove that comment or update it to say toggle since thats what its doing

CMakeLists.txt
```diff
@@ -84,6 +84,17 @@ set_target_properties(relive_api      PROPERTIES UNITY_BUILD OFF) # too big
 set_target_properties(relive_api_test PROPERTIES UNITY_BUILD OFF) # too big
 # -----------------------------------------------------------------------------
 
+# -----------------------------------------------------------------------------
+# Setup GCC and Clang warnings
+add_library(project_warnings INTERFACE)
+export(TARGETS project_warnings FILE project_warnings.cmake)
+
+if(NOT MSVC)
```
paulsapps, line 92: would we not want to kill the manual MSVC warning stuff I added and use this instead?

Source/AliveLibAE/BirdPortal.cpp
paulsapps: seems that apparently it should fall through according to mouze

SuperV1234: Will fix

Source/AliveLibAE/NakedSlig.cpp
SuperV1234: Will add TODO

Source/AliveLibAO/Input.cpp
SuperV1234: Will fix

Source/AliveLibAO/MainMenu.cpp
SuperV1234: Will change

PR #1270, opened 2021-04-26
`relive_api` micro-optimizations

SuperV1234:
Shaves 9-10s off `relive_test_api` on my machine, on a debug build


Source/Tools/relive_api/relive_api_test.cpp
```diff
@@ -81,37 +81,56 @@ template <typename... Ts>
     return concat(kAODir, fileName);
 }
 
+[[nodiscard]] std::vector<u8>& getStaticFileBuffer()
+{
+    static std::vector<u8> result;
+    return result;
+}
+
+[[nodiscard]] std::vector<u8>& readFileIntoStaticFileBuffer(const std::string& filePath)
```
paulsapps, line 90: does this really make a huge change to performance? These static buffers make adding threading later harder can sometimes cause issues since the buffer can contain unexpected data

Source/Tools/relive_api/relive_api.hpp
```diff
@@ -98,9 +98,16 @@ namespace ReliveAPI
     void DebugDumpTlvs(const std::string& prefix, const std::string& lvlFile, s32 pathId);
 
     API_EXPORT [[nodiscard]] s32 GetApiVersion();
+
+    API_EXPORT void ExportPathBinaryToJson(std::vector<u8>& fileDataBuffer, const std::string& jsonOutputFile, const std::string& inputLvlFile, s32 pathResourceId);
```
paulsapps, line 102: if `std::vector<u8>& fileDataBuffer` was to speed up the tests then I think there needs to be an overload that doesn't have this argument as the things calling this API will now need updating and its less obvious what is happening

SuperV1234: There is, is on line 103 :)

Source/Tools/relive_api/relive_api_test.cpp
SuperV1234: It contributed to the speedup I have seen. We can change `static` to `thread_local` if we add threading in the future.

PR #1272, opened 2021-04-27
Fixes Death Delay Glitch

deanff:
Tested by doing the following as fast as Abe can:
1. Throw grenade down
2. Jump up
3. Sneak once
4. Hold hop
This used to always result in DDG, but not anymore.


Source/AliveLibAO/Abe.cpp
```diff
@@ -1125,6 +1125,9 @@ void Abe::vUpdate_41FDB0()
                     field_112_prev_motion = 0;
                     field_108_bMotionChanged = 1;
                     field_106_shot = 0;
+#if ORIGINAL_GAME_FIXES || ORIGINAL_GAME_FIX_DEATH_DELAY
+                    field_2A8_flags.Clear(Flags_2A8::e2A8_Bit2); // OG Change - Fixes "Death Delay Glitch"
```
paulsapps, line 1129: Is the meaning of known? If so can we rename it whatever it actually means? Also does this bug not happen in AE?

deanff: Do you mean the meaning of "e2A8_Bit2"? 
It seems to be the equivalent of "e1AC_Bit2_return_to_previous_motion" in AE.
Would you like me to rename all instances of "e2A8_Bit2" to "e2A8_Bit2_return_to_previous_motion"?
The bug can happen in AE: https://www.youtube.com/watch?v=GXlXGiCgI_M
But I'm not sure how it is triggered... Might be with the electric fences.

paulsapps: yes renaming to e2A8_Bit2_return_to_previous_motion would be good, since it can happen in AE but isn't fixed there you should change ORIGINAL_GAME_FIX_DEATH_DELAY to be ORIGINAL_GAME_FIX_DEATH_DELAY_AO too for now

Issue #1171, opened 2021-03-17
AO OG Bug - Speak I+II can be remapped when using the keyboard


pryon:
**Description of issue:** 
The Speak I and Speak II inputs can be remapped to basically any key in Oddysee on the keyboard remapping screen.
These inputs were only meant to be used by controllers, not keyboards, which is why they cannot even be selected for remapping in Exoddus when using the keyboard.

For keyboards, the keys <kbd>1</kbd> to <kbd>8</kbd> are to be used for gamespeak.

**Steps to repro:** 
1. Select the keyboard as the input device
2. Remap Speak I and Speak II to anything you want

PR #1274, opened 2021-04-28
Prevents users from editing Speak1 and Speak2 on Keyboard

deanff:
Fixes issue #1171

### DIFF
--- a/Source/AliveLibAO/MainMenu.cpp
+++ b/Source/AliveLibAO/MainMenu.cpp
@@ -3269,12 +3269,36 @@ void Menu::ButtonRemap_Update_47F6F0()
         field_1E0_selected_index--;
         SFX_Play_43AE60(SoundEffect::MenuNavigation_61, 45, 400, 0); // TODO: Input constants
         bWaitingForRemapInput_9F2DE8 = 1;
+
+#if ORIGINAL_PS1_BEHAVIOR
+        if (!Input_JoyStickEnabled())
+        {
+            if (field_1E0_selected_index < 0)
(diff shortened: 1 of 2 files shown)


Source/AliveLibAO/MainMenu.cpp
```diff
@@ -3269,12 +3269,36 @@ void Menu::ButtonRemap_Update_47F6F0()
         field_1E0_selected_index--;
         SFX_Play_43AE60(SoundEffect::MenuNavigation_61, 45, 400, 0); // TODO: Input constants
         bWaitingForRemapInput_9F2DE8 = 1;
+
+#if ORIGINAL_PS1_BEHAVIOR
+        if (!Input_JoyStickEnabled())
+        {
+            if (field_1E0_selected_index < 0)
+            {
+                field_1E0_selected_index = (int)RemapOption::eCrouch_6;
```
paulsapps, line 3278: seems like field_1E0_selected_index needs changing to be of type RemapOption? That will ditch these casts, if we do need a cast it should be a C++ cast (static_cast<T>(a)) in this case.

deanff: field_1E0_selected_index is used for ALL main menu selections, not just the controls remap page.
namespace enums ditch the cast and won't clutter the namespace. What are your thoughts on them?

paulsapps: What about using a union with an enum class and a raw type inside of it like the Abe.hpp?

PR #1283, opened 2021-05-01
Force CWD on macOS to be the folder containing the executable

LIJI32:


Source/relive/Exe.cpp
```diff
@@ -257,3 +257,21 @@ bool RunningAsInjectedDll()
 {
     return false;
 }
+
+#ifdef __APPLE__
+#include <mach-o/dyld.h>
+
+static void __attribute__((constructor)) FixCWD()
+{
+    char* imagePath = strdup(_dyld_get_image_name(0));
```
paulsapps, line 266: shouldn't this be checked for returning nullptr?

LIJI32: No, it should never return `NULL` (if it does, the dynamic loader is pretty much dead and we're going to crash anyway)

paulsapps: no I mean strdup

LIJI32: Ah. Well kind of the same applies, if one of the very first `malloc`s we perform fails we're not going to be able to do anything useful anyway. I'm generally against checking for `malloc` failures since they're almost always irrecoverable, but your call.

paulsapps: eh it will keep static analysis tools happy I guess?

LIJI32: They actually check for these? I haven't seen code testing for `malloc`'s return value for *ages*. :|

paulsapps: they get flagged as security exploits

PR #1298, opened 2021-05-20
Fixed wrong char type in debug code.

Leonard2:
Hey, first pr!
Just testing the waters.
This fixes compilation for me on MSVC 2019, Visual studio just _refuses_ to take any signed char pointers for common string types.

By the way, how do you guys normally enable developer_mode?
I had to force it "ON" in options.cmake because it would override the header file at build time but then it shows in git status.


paulsapps:
the dev mode is a bit of a hack and the pause menus of both games really need a rewrite and merging into 1/making easily extendable

PR #1299, opened 2021-05-21
Automatically switches between keyboard/gamepad using SDL Events - Joystick Added/Removed

deanff:
Automatically switches between keyboard and gamepad by detecting what is connected via SDL Events. Also allows player to configure gamepad in Abe's Exoddus without having to press C on the keyboard.


Source/AliveLibAE/Sys.cpp
```diff
@@ -1022,7 +1023,30 @@ EXPORT s8 CC Sys_PumpMessages_4EE4F4()
...
+        {
+            totalConnectedJoysticks++;
+            SDL_Log("User just inserted joystick!");
+            Input_Init_491BC0();
+            sJoystickEnabled_5C9F70 = 1;
+        }
+        else if (event.type == SDL_JOYDEVICEREMOVED)
+        {
+            totalConnectedJoysticks--;
+            SDL_Log("User just removed joystick!");
```
paulsapps, line 1037: should prob use LOG_INFO like all the other logging, ditto for the one above

Source/AliveLibAO/Map.hpp
```diff
@@ -39,6 +39,28 @@ struct Map_PathsArray final
 };
 ALIVE_ASSERT_SIZEOF(Map_PathsArray, 120);
 
+namespace CameraIds
```
paulsapps, line 42: you can use CameraIds::Menu if you like, I don't mind which one is used

Source/AliveLibAE/Sys.cpp
```diff
@@ -1022,7 +1023,30 @@ EXPORT s8 CC Sys_PumpMessages_4EE4F4()
     SDL_Event event;
     while (SDL_PollEvent(&event))
     {
-        // Pump events
+#if ORIGINAL_PS1_BEHAVIOR // OG Change - Automatically determines active controller (gamepad/keyboard)
```
paulsapps, line 1026: I don't think this should all be using ORIGINAL_PS1_BEHAVIOR? Probably another option called auto switch to game pad or something?

Source/AliveLibAE/MainMenu.cpp
```diff
@@ -2696,7 +2696,13 @@ MainMenuNextCam MainMenuController::ControllerMenu_Update_4D16D0(u32 input)
         return MainMenuNextCam(MainMenuCams::eOptionsCam);
     }
 
-    if (input & InputCommands::Enum::eConfigure)
+#if ORIGINAL_PS1_BEHAVIOR // OG Change - Allows player to select "Configure Controller" using their gamepad (Square/Circle Button)
```
paulsapps, line 2699: shouldn't this be AUTO_SWITCH_CONTROLLER as well?

deanff: It's unrelated. I'll create a new cmake option for this one too.

PR #1306, opened 2021-05-31
FMV Audio Fix

deanff:


paulsapps:
does this not de-sync then? it seems to be tied to real time rather than the audio clock?

deanff:
Haven't had issues. There is one thing that's a bit different though... If you grab the window and move it around, the game will get in a temporary audio loop until you let go. AE does this as well. What's different is AO will catch up! Whether this is an issue - I'm unsure, was kinda hoping that we could somehow fix the whole temp audio loop thing and just have normal playback when dragging the window.

paulsapps:
Should we remove the code that disables the FMVs if they now work then?

deanff:
I think so... The only remaining (minor) issue that springs to mind is the colours - the ends of FMVs and the start of Gameplay / the main menu don't always match up, and vice versa.

paulsapps:
I think that is an OG issue

deanff:
Yeah - I think you're right. PS1 has matching colors, but OG PC port does not.

paulsapps:
You gonna remove the if 0 then?

deanff:
Which if 0? I've removed the #if 1 with the comment "Don't play FMVs for now till sound issues are fixed"

paulsapps:
ah yes, I am the blind mud

paulsapps:
So after some testing I do see one bug, the audio isn't stopped so you can hear music at the same time

PR #1307, opened 2021-05-31
Stop Background Audio to Play FMV Audio

deanff:


Source/AliveLibAO/Movie.cpp
```diff
@@ -152,6 +152,8 @@ class PsxStr
                     // If this is the first time then start to play the buffer
                     if (!bStartedPlayingSound && !bNoAudioOrAudioError)
                     {
+                        SND_StopAll_4CB060();
```
paulsapps, line 155: do we not need a call to restart audio at the end like the pause menu?

deanff: I don't think we need it. It's needed in the pause menu to make background music start up again, but it seems like music will start up anyway at the end of the FMV / start of gameplay.
That said, I'm having some doubts about this commit... There appears to be areas in the PS1 version where the background audio continues over the FMV... https://youtu.be/yW5IRvW4apo?t=4958
Thoughts? I might load a save state tomorrow and test some more...

paulsapps: maybe comment it with a TODO - there might be a field in the FMV structure that tells you if the music should continue or not, actually that is very likely thinking about it

paulsapps: I don't think it does happen ever in PC tho - prob another OG bug

deanff: Yeah, it looks like the OG PC Port would cull the audio for all FMV transitions:
https://youtu.be/1wx7a_RAqq4?t=1251
Going to leave a TODO comment.

paulsapps: yes - but it does restart it after, which we currently don't

deanff: What makes you say it needs a restart?
Abe's Exoddus doesn't call SND_Restart after its FMVs.
I added one to AO and was unable to notice any difference.

I did, however, notice another sound-related "bug" (or at least a difference between relive and the OG PC port)...

OG PC Port: Background audio starts just before FMV. Does not restart after FMV.
https://youtu.be/1wx7a_RAqq4?t=1250

Relive always restarts the bg audio - With or without SND_Restart. 
You can go in the door between upper/lower rupture farms over and over again - and the bg music will always replay what was playing before.
This is not the case for OC PC Port or PS1 version.

paulsapps: hmm I'm not sure what you mean, should probably raise a bug to cover these, I'll merge this PR then

PR #1318, opened 2021-07-02
don't try to place elum if hes not following (fix #1304)

### LINKED ISSUE #1304: AO Rev Bug - Elum teleports to the new path when abe does a path transition while elum follows him
Steps to repro:
1. load the save
2. tell elum to follow you
3. go two screens to the right to do a path transition
4. go back

https://user-images.githubusercontent.com/43358824/124368788-1100ba00-dc65-11eb-9afa-0e1742b14b14.mp4

save: [!.zip](https://github.com/AliveTeam/alive_reversing/files/6759160/default.zip)

function: ``Map::Handle_PathTransition_444DD0``

paulsapps:


paulsapps:
not merging as there is still a teleporting bug to fix :|

PR #1335, opened 2021-08-18
add and name all AE Menu `AnimdId`'s, use some `AnimRecord`'s and name/rename some stuff

mouzedrift:


Source/AliveLibAE/MainMenu.cpp
```diff
@@ -3398,15 +3419,15 @@ s32 MainMenuController::ChangeScreenAndIntroLogic_4CF640()
 
             field_20_animation.field_4_flags.Clear(AnimFlags::eBit3_Render);
             field_20_animation.field_4_flags.Clear(AnimFlags::eBit2_Animate);
-            field_21E_bChangeScreen = 3;
+            field_21E_changeScreenState = 3;
```
paulsapps, line 3422: seems you could have used CameraSwapEffects for these :TapeDan:

PR #1358, opened 2021-11-29
display a warning when a switch id exceeds the max array size

mouzedrift:
id's in the new qt editor should probably be capped to 256 as well since everything above it screws up things


paulsapps:
makes sense, but shouldn't we also change the json type range so you also can't use a value over 255?

PR #1367, opened 2021-12-05
Remove busy loops

ilelann:
replaced with either
    this_thread::sleep_*
    condition_variable for audio read/write loop

this reduces usage from "around 16%" (on a recent 16t CPU, thus a 2.5 threads load)
to a nicer "around 0.7%"

### DIFF
--- a/Source/AliveLibAE/Sound/PsxSpuApi.cpp
+++ b/Source/AliveLibAE/Sound/PsxSpuApi.cpp
@@ -9,6 +9,7 @@
 #include "PathData.hpp" // SoundBlockInfo, SeqPathDataRecord
 #include "../AliveLibAE/Io.hpp"
 #include <assert.h>
+#include <thread>
 
 struct VagAtr final
 {
@@ -1346,9 +1347,40 @@ EXPORT void CC SsSetTickMode_4FDC20(s32)
     // Stub
 }
 
+EXPORT bool CC SsSeqCalledTbyT_4FDC80_sleep_until_due_time_if_before_max_time(SYS_time_point_t current_time, SYS_time_point_t max_time)
+{
+    if (!gSpuVars->sbDisableSeqs())
+    {
+        const auto lastTime = gSpuVars->sLastTime();
+        if (lastTime == 0xFFFFFFFF)
+        {
+            return true;
+        }
+        else
+        {
+            const auto nextDueTime = lastTime + 30;
+
+            if ((s32) (SYS_AsU32Ms(max_time) - nextDueTime) < 0)
+            {
+                // we won't make it in time, skip
+                return false;
+            }
+
+            const auto timeToWait = (s32) (nextDueTime - SYS_AsU32Ms(current_time));
+            if (timeToWait <= 0)
+            {
+                // we're late or right on time, hurry!
+                return true;
+            }
+
+            std::this_thread::sleep_for(std::chrono::milliseconds{timeToWait});
+            return true;
+        }
+    }
+    return false;
+}
+
 // TODO: Removed 4FDC30
-
-
 EXPORT void CC SsSeqCalledTbyT_4FDC80()
 {
     if (!gSpuVars->sbDisableSeqs())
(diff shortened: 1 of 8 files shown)


ilelann:
Some missing include <thread> for Movie.cpp is killing linux/macos builds. I will fix that.

I probably also need to replicate AO's Movie.cpp fixes in AE.
I only tested AO but I will test AE soon.

Still interested in any feedback meanwhile.

Source/AliveLibAE/Sound/PsxSpuApi.cpp
```diff
@@ -1346,19 +1347,51 @@ EXPORT void CC SsSetTickMode_4FDC20(s32)
     // Stub
 }
 
-// TODO: Removed 4FDC30
+EXPORT bool CC SsSeqCalledTbyT_4FDC80_sleep_until_due_time_if_before_max_time(SYS_time_point_t current_time, SYS_time_point_t max_time)
```
paulsapps, line 1350: _4FDC80 needs to be at the end of the function as exported function names are parsed at runtime for installing hooks into the game when running as a DLL. This name is super verbose too, I would have just left it as SsSeqCalledTbyT_4FDC80 tbh since the waiting/sleeping is an implementation detail.

Source/AliveLibAE/Sound/PsxSpuApi.cpp
```diff
@@ -1346,19 +1347,51 @@ EXPORT void CC SsSetTickMode_4FDC20(s32)
...
+        const auto lastTime = gSpuVars->sLastTime();
+        if (lastTime == 0xFFFFFFFF)
+        {
+            return true;
+        }
+        else
+        {
+            const auto nextDueTime = lastTime + 30;
+
+            if ((s32) (SYS_AsU32Ms(max_time) - nextDueTime) < 0)
```
paulsapps, line 1363: use static_cast instead of a C cast

Source/AliveLibAE/Sound/PsxSpuApi.cpp
```diff
@@ -1346,19 +1347,51 @@ EXPORT void CC SsSetTickMode_4FDC20(s32)
...
+            {
+                // we won't make it in time, skip
+                return false;
+            }
 
+            const auto timeToWait = (s32) (nextDueTime - SYS_AsU32Ms(current_time));
```
paulsapps, line 1369: use static_cast instead of a C cast

Source/AliveLibAE/Sound/SDLSoundSystem.cpp
```diff
@@ -100,6 +100,11 @@ HRESULT SDLSoundSystem::Release()
 
         // Stop audio rendering thread
         mRenderAudioThreadQuit = true;
+        {
+            std::lock_guard g{mAudioRingBufferMutex};
```
paulsapps, line 104: instead of 'g' can we at least called it 'lock' or something

Source/AliveLibAE/Sound/SDLSoundSystem.cpp
```diff
@@ -139,6 +144,11 @@ void SDLSoundSystem::AudioCallBack(Uint8* stream, s32 len)
     {
         LOG_ERROR("Ring buffer read failure!");
     }
+
+    {
+        std::lock_guard g{mAudioRingBufferMutex};
```
paulsapps, line 149: ditto for naming

Source/AliveLibCommon/Sys_common.cpp
```diff
@@ -13,16 +13,19 @@
...
 }
 
+SYS_time_point_t SYS_GetTime()
+{
+    return std::chrono::steady_clock::now();
+}
+
+u32 SYS_AsU32Ms(SYS_time_point_t t)
+{
+    return (u32) std::chrono::duration_cast<std::chrono::milliseconds>(t.time_since_epoch()).count();
```
paulsapps, line 23: static_cast instead of C cast

paulsapps:
We could probably do with someone with a slower machine testing this to check it doesn't impact performance. Maybe @mouzedrift  can try one the build is fixed.

mouzedrift:
CPU usage went down to +-7% but the game looks very laggy

paulsapps:
I suspected that might be an issue. @ilelann it might be better to try to get https://github.com/AliveTeam/sound_rev working instead as this can work without another thread which would remove the busy waits. It needs progressing so that the emulator integration can run the psx SPU code (instead of emulated mips code). Once this works it can be refactored/reduced down and integrated into relive.

ilelann:
Thanks a lot for your comments, I'll dig that.
@mouzedrift  May I know what kind of hardware & OS was used for your test ?

mouzedrift:
windows 10 pro, 8 gb ram, intel core i5-4440 with intel hd graphics 4600

Source/AliveLibAE/Sound/PsxSpuApi.cpp
ilelann: I meant to avoid breaking existing SsSeqCalledTbyT_4FDC80 ABI with "max_time" new argument.
(because I understood that number-suffixed functions are meant to be interchangeable with original binaries)
Unless I miss something, I thus can simplify the name but not use the original one.

paulsapps:
A temp hack fix was merged that removes the busy loops, already confirmed that this causes stuttering on some machines but it can be turned off via the ini. Hack fix is temp till audio is rewritten.

PR #1383, opened 2021-12-16
Small refactoring and de-duplication of code

gapys-krzysztof:


Source/AliveLibAE/Renderer/IRenderer.cpp
```diff
@@ -24,9 +27,11 @@ void IRenderer::CreateRenderer(Renderers type)
             gRenderer = new SoftwareRenderer();
             break;
 
+#if RENDERER_OPENGL
```
paulsapps, line 30: the idea was to remove as many of these as possible rather than add more ;)

PR #1384, opened 2021-12-16
Coupling Oddysee & Exoddus Classes - Proof of Concept

gapys-krzysztof:
Though rather crude, seems to be doing the trick so far.

Please note the comment about Options::eCantKill_Bit11 clearing before merging. If there's no concern, I'd like to remove both comments.


paulsapps:
This won't work as it breaks ABI - merging will require breaking ABI and at that point we'd simply just have one BaseGameObject class that encapsulates the functionality of both games. We need the automated demo record/play back and verification implemented before this for automated regression testing.

gapys-krzysztof:
> at that point we'd simply just have one BaseGameObject class that encapsulates the functionality of both games.
To be honest, this and the closest derivative classes are rather big and/or a dependency for most classes in the games, so I don't see them wholy coupled together in one go.

Not sure how I would tackle automated demo record/playback. I initially thought of setting up a GTest with a sandbox level where you can predefine events like AliveObject spawning and interactions, but the process of getting there is rather tough.

paulsapps:
`To be honest, this and the closest derivative classes are rather big and/or a dependency for most classes in the games, so I don't see them wholy coupled together in one go.`

Not sure what you mean by this.

The demo recording is when you leave the game idle and automated game play starts. But we need to extend this to also save/restore random seeds because it isn't deterministic. And we need to save/verify the position/state of objects to check the playback is still in sync with what was recorded.

gapys-krzysztof:
>Not sure what you mean by this.

Just that the code is connected to a lot of other code areas, and I'd probably need to retest everything after each function change. So, yeah. Demo playback should come first.

>The demo recording is when you leave the game idle and automated game play starts. But we need to extend this to also save/restore random seeds because it isn't deterministic. And we need to save/verify the position/state of objects to check the playback is still in sync with what was recorded.

I'll look into storing random seeds.

paulsapps:
It might be worth implementing fixed quick save for AE and AO as part of this. For the state sync check you could just do a quick save and check its equal to the quick save in the recording every X amount of frames.

That way we could allow record and playback from any point in the game too. Only things to consider is loading/FMV playback shouldn't count as a recorded frame to avoid a desync.

Would prob want to allow the playback to happen from the command line too so we can have a script that runs multiple instances of the game in parallel to test full game coverage in as little time as possible.

PR #1385, opened 2021-12-16
Demo order fix.

gapys-krzysztof:
The variable responsible for storing ID of the currently playing demo was incremented too quickly, therefore the first demo  was skipped when shouldn't.

Issue was only present in RELIVE

### DIFF
--- a/Source/AliveLibAE/MainMenu.cpp
+++ b/Source/AliveLibAE/MainMenu.cpp
@@ -2207,6 +2207,14 @@ MainMenuNextCam MainMenuController::LoadDemo_Update_4D1040(u32)
         }
         field_6_flags.Set(Options::eDead_Bit3);
 
+
+        char_type file[32] = {};
+        sprintf(file, "ATTR%04d.SAV", sDemos_5617F0[demoId].field_A_id);
+        ResourceManager::LoadResourceFile_49C170(file, nullptr);
+        u8** resource = ResourceManager::GetLoadedResource_49C2A0(ResourceManager::ResourceType::Resource_NxtP, ResourceID::kUnknownResID_0, 1, 0);
+        sActiveQuicksaveData_BAF7F8 = *(reinterpret_cast<Quicksave*>(*resource));
+        ResourceManager::FreeResource_49C330(resource);
+
         demoId = sDemoIdChosenFromDemoMenu_5C1B9E;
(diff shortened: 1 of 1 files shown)


Source/AliveLibAE/MainMenu.cpp
```diff
@@ -2207,6 +2207,13 @@ MainMenuNextCam MainMenuController::LoadDemo_Update_4D1040(u32)
         }
         field_6_flags.Set(Options::eDead_Bit3);
 
+        char_type file[32] = {};
+        sprintf(file, "ATTR%04d.SAV", sDemos_5617F0[demoId].field_A_id);
```
paulsapps, line 2211: I don't see how this is correct when the code below in !gIsDemoStartedManually_5C1B9C changes the value of demoId

gapys-krzysztof: It changes it for the purpose of having it ready for the next demo

gapys-krzysztof: To illustrate it better, I didn't move this block of the code, but moved the one that increments sCurrentDemoIdForIdlingDemoPlayback_5C1BA2 under it.

paulsapps: But `demoId` is a local, and after `demoId = sCurrentDemoIdForIdlingDemoPlayback_5C1BA2;` its never used again. But it was used before you moved the code hence it seems wrong or needs more refactoring.

gapys-krzysztof: You're right, demoId isn't used later on. Since I don't see why it should be changed right before loading a savegame (wouldn't make much sense, since it checks if the demo exists earlier on, to pick a different one later), so I removed those assignments. They point to the same variables anyway (albeit not needlessly incremented)

paulsapps: But it was used before the change and it doesn't point to the same var - is this fixed based on the original func by comparing with IDA?

gapys-krzysztof: No, I don't have IDA. The fix was only based on observation that the original game starts demos from the first one forward, and this project from the second one forward

paulsapps:
waiting for @mouzedrift to cross ref with original func

PR #1466, opened 2022-04-16
Audio panning

sean-horton:
2 things done:

- These changes allow audio panning of audio sequences, most notable on the title tracks of both AO and AE. 
- I commented "wait until" logic around the sequence loop. It seems to have waited every 30ms before checking sequences. While it's only 30ms, audio sounded like it was racing or slowing down at times, i.e. off beat. I wonder if this check was originally to save CPU cycles, but I doubt it's an issue on new hardware.

There seemed to be extra logic with 64 (center) and 127 (right) values, I commented them out instead of removing them,  unsure how you want to handle it, for original reversing efforts. 

As a side note there was actually a comment in `PsxSpuApi.cpp` about assigning `field_3_pan` to `field_11_pad`. I'm tempted to change `field_11_pad` to `pan`.

Further, the change is duplicated for AO and AE, combining them would be too large of a change for this PR I think. Also, some of the values passed in places for left or right pan are unused now, so cleanup could be done.

### DIFF
--- a/Source/AliveLibAE/Abe.cpp
+++ b/Source/AliveLibAE/Abe.cpp
@@ -587,7 +587,7 @@ EXPORT s32 CC Environment_SFX_457A40(EnvironmentSfx sfxId, s32 volume, s32 pitch
                     static_cast<s16>(2 * sndVolume / 9),
                     static_cast<s16>(2 * sndVolume / 9),
                     static_cast<s16>(pitchMin),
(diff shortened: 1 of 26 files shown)


sean-horton:
Closing this pull request as it is the wrong approach. I opened a second pull request that may be better.
https://github.com/AliveTeam/alive_reversing/pull/1469

PR #1486, opened 2022-05-05
Fix bug #1342 - movie transition loop when riding elum

### LINKED ISSUE #1342: AO Rev Bug - You can easily get stuck in a movie transition loop at the beginning of scrabania
Steps to repro:
1. load the save
2. mount elum and walk to the left until you can't see elum/abe anymore
3. hold shift and tap the <kbd><</kbd> key once

you're supposed to stay in the next camera after the transition but in the rev version you end up going back

save:
[!.zip](https://github.com/AliveTeam/alive_reversing/files/7028414/default.zip)

## rev

https://user-images.githubusercontent.com/43358824/130375466-7336dbde-7d71-4cb9-a7db-d2cd1e606e3e.mp4

## OG

https://user-images.githubusercontent.com/43358824/130375472-08094832-7b00-41c1-ac18-04c6e7dd5731.mp4

Retoxified:
Fixes #1342 - AO Rev Bug - You can easily get stuck in a movie transition loop at the beginning of scrabania

### DIFF
--- a/Source/AliveLibAO/Elum.cpp
+++ b/Source/AliveLibAO/Elum.cpp
@@ -837,33 +837,30 @@ void Elum::HandleElumPathTrans_411460()
         field_AC_ypos = FP_FromInteger(camCoords.field_2_y + FP_GetExponent(field_AC_ypos) % 480);
     }
 
-    if (field_F4_pLine)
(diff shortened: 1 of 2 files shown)


Retoxified:
Also solves #1304 AO Rev Bug - Elum teleports to the new path when abe does a path transition while elum follows him

PR #1492, opened 2022-05-11
Fix for gamespeak breaking after quitting demo mid-hello

### LINKED ISSUE #907: AE OG Bug - Interrupting demo during Abe gamespeak breaks gamespeak & causes special case crashes
**Description of issue:** 
If a currently playing demo gets aborted by hitting any key while Abe is talking, the next time the demo runs, gamespeak will stop eliciting responses from mudokons, as well as breaking demo movement. If instead of restarting the demo, the game is played instead, the same can be observed.

In special cases a crash can also be triggered, e.g. by aborting and restarting the Blind Mudokons 2 demo.

**Steps to repro:** 
1. Navigate to Options -> Demos in the main menu
2. Start playing *Blind Mudokons 2*
3. Abort demo while Abe is saying hello
4. Restart demo, wait a little until Abe says hello again

Interestingly enough this scenario does not crash R.E.L.I.V.E., but causes graphical glitches:
image

Retoxified:
Fixes #907 AE OG Bug - Interrupting demo during Abe gamespeak breaks gamespeak & causes special case crashes by unsetting sIsMudStandingUp_5C3018 when a demo is ended by a keypress.

### DIFF
--- a/Source/AliveLibAE/QuikSave.cpp
+++ b/Source/AliveLibAE/QuikSave.cpp
@@ -342,6 +342,10 @@ EXPORT void CC Quicksave_LoadFromMemory_4C95A0(Quicksave* quicksaveData)
         0,
         1);
     gMap_5C3030.field_8_force_load = 1;
+
+#if ORIGINAL_GAME_FIXES
+    sIsMudStandingUp_5C3018 = 0;
+#endif // ORIGINAL_GAME_FIXES
 }
 
 EXPORT void CC Quicksave_LoadActive_4C9170()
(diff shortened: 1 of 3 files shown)


Source/AliveLibAE/QuikSave.cpp
```diff
@@ -342,6 +342,10 @@ EXPORT void CC Quicksave_LoadFromMemory_4C95A0(Quicksave* quicksaveData)
         0,
         1);
     gMap_5C3030.field_8_force_load = 1;
+
+#if ORIGINAL_GAME_FIXES
+    sIsMudStandingUp_5C3018 = 0;
```
paulsapps, line 347: should this not actually be part of the quick save data structure? it seems like something that should be persisted?

Retoxified: Ideally, probably yes, it didn't seem to have any negative side effects to do it this way during my tests, and I wasn't sure if I could modify quicksave/quickload data without messing with actual save files as well, as it seemed to read the quicksave data from the save file if I'm not mistaken

PR #1520, opened 2022-05-28
fix #1511

### LINKED ISSUE #1511: AO Rev Bug - Pause menu level/path/camera info on linux isn't centered
image

mouzedrift:


mouzedrift:
seems like this fix doesn't work so i need to look at this again

PR #1601, opened 2023-05-21
Audio SPU emulation (duckstation)

sean-horton:
This is an attempt at replacing sound with Duckstation's SPU. It is a non-ABI breaking change and can be turned off and on using `options.cmake` value `AUDIO_SPU_EMULATION`. 

I have not hooked up FMV audio to the SPU emulator, but it should be trivial. I'll wait to see if this PR is even something that would be wanted. 

- `MidiPlayer` defines the SPU/SND functions. 
- `Sequencer` is the modified duckstation SPU 
- `oddio` defines fixes for vag attributes (which I'll mention more below)

The SPU runs at 44100hz which is executing in the SDL callback. The emulator should probably run in its own thread as running calculations in the SDL callback is probably bad practice. Unsure if this will cause clipping on slower CPU's. CPU usage seems comparable to master branch, however if many sounds are playing (slig constantly shooting), my CPU usage is slightly higher than master. 

Interestingly the PC vag attributes differ from PSX. This can be seen by disabling the vag attribute fixes by commenting out `mSoundSampleParser.applyFix(...)` in `MidiPlayer`. With it commented it will sound like pc, with it uncommented it will sound like psx. One obvious example is Abe's "Hello". He has a deeper voice in pc, whereas psx is slightly higher pitch. 

R6 is even more messed up and fixes can not be applied without more inspection, right now it uses all pc sounds, so Abe will have pc's lower pitch voice.

Further, I believe a lot of the volumes are incorrect in the sound definitions within things like `slig.cpp`. 
1. When a slig stops shooting it plays `eHitBottomOfDDeathPit` at full volume. This should almost certainly be played at a lower volume. `SND_SEQ_SetVol()`
2. The sequence played for doves in, for example, paramonia, play at full volume, and should have `SND_SEQ_SetVol()`
3. And; Leaves shooting out of wells seem to have too low of volume
4. Probably others...

I have not applied fixes for AE, so that could probably be improved. However, I know greeters wheeling around make no noise, but it's because their volume is too low, increasing it the sound will play.

In short; I'm pretty sure the pc port introduced more sound issues than just the player, most likely to cover for the incorrect player.


Source/AliveLibAE/Sound/Midi.cpp
```diff
@@ -867,3 +867,220 @@ EXPORT void CC SND_Restart_4CB0E0()
...
+    sounds.dat file whereas AO reads it inline of the vab body.
+*/
+class AESoundSampleParser : public psx::SoundSampleParser
+{
+public:
+    std::vector<psx::Sample*> parseSamples(psx::VabHeader* vabHeader, u8* ppVabBody)
+    {
+        std::vector<psx::Sample*> samples;
+        
+
```
d3nwah, line 942: Double endline

Source/AliveLibAE/Sound/Midi.cpp
```diff
@@ -867,3 +867,220 @@ EXPORT void CC SND_Restart_4CB0E0()
...
+            throw std::invalid_argument("Could not find sounds.dat");
+        }
+        soundDatFile.seekg(0, std::ios::end);
+        std::streamsize fileSize = soundDatFile.tellg();
+        soundDatFile.seekg(0, std::ios::beg);
+        char* soundData = new char[(s32) fileSize + 1]; // Plus one, just in case interpolating tries to go that one byte further!
+
+        if (soundDatFile.read(soundData, fileSize))
+        {
+            // Tada!
```
d3nwah, line 956: 👀
I don't see anything happening here, what's the magic trick?

Source/AliveLibAE/Sound/Midi.cpp
```diff
@@ -867,3 +867,220 @@ EXPORT void CC SND_Restart_4CB0E0()
...
+}
+
+EXPORT void CC SND_Shutdown_4CA280()
+{
+    player->SND_Shutdown();
+}
+
+EXPORT void CC SND_Stop_Channels_Mask_4CA810(u32 bitMask)
+{
+    bitMask;
```
d3nwah, line 1012: This line, and the ones in SND_Load_VABS_4CA350 don't seem to do anything, is this needed?

Source/AliveLibAE/Sound/Midi.cpp
```diff
@@ -867,3 +867,220 @@ EXPORT void CC SND_Restart_4CB0E0()
...
+{
+    return player->SFX_SfxDefinition_Play(reinterpret_cast<psx::SfxDefinition*>(const_cast<SfxDefinition*>(sfxDef)), volLeft, volRight, pitch_min, pitch_max);
+}
+
+EXPORT s32 CC SFX_SfxDefinition_Play_4CA420(const SfxDefinition* sfxDef, s16 volume, s16 pitch_min, s16 pitch_max)
+{
+    return player->SFX_SfxDefinition_Play(reinterpret_cast<psx::SfxDefinition*>(const_cast<SfxDefinition*>(sfxDef)), volume, pitch_min, pitch_max);
+} 
+
+#endif
```
d3nwah, line 1086: It's good practice to end the source code files with an endline at the end.  Also, not having one in an .hpp file can cause problems with including multiple ones one after the other.

Seems multiple files miss it here.

Source/AliveLibAE/Sound/Sound.cpp
```diff
@@ -228,8 +228,39 @@ EXPORT s32 CC SND_PlayEx_4EF740(const SoundEntry* pSnd, s32 panLeft, s32 panRigh
     pDSoundBuffer->SetFrequency(freqHz);
     pDSoundBuffer->SetVolume(sVolumeTable_BBBD38[panRightConverted]);
 
+    // OLD PAN
```
d3nwah, line 231: If the new panning is superior, no need to keep the original bad one in, should be removed instead, with all vars that it exclusively uses

Source/AliveLibCommon/CMakeLists.txt
```diff
@@ -35,7 +35,13 @@ SET(AliveLibSrcCommon
...
     Psx_common.hpp
     W32CrashHandler.hpp
-)
+    audio/Sequencer.hpp
+    audio/Sequencer.cpp
+    audio/Stream.hpp
+    audio/Stream.cpp
+    audio/MidiPlayer.hpp
+    audio/MidiPlayer.cpp
+    audio/oddio.h)
```
d3nwah, line 44: Let's keep ) on the new line like it was originally :)

Source/AliveLibCommon/audio/MidiPlayer.hpp
```diff
@@ -0,0 +1,297 @@
...
+        /* Read a file within the game, example a .VB file */
+        virtual ResourceData* readFile(char_type* name)
+        {
+            name;
+            throw std::invalid_argument("readFile is not overridden");
+        }
+        /* AO reads sequence inline with data. AE reads from a sounds.dat file */
+        virtual ResourceData* readSeq(const char_type* fileName, const char_type* sequenceName)
+        { 
+            fileName;
```
d3nwah, line 217: Probably safe to remove, like the other lines like this in this file (and others)

Source/AliveLibCommon/audio/Sequencer.cpp
```diff
@@ -0,0 +1,1552 @@
...
+{
+    mutex.lock();
+    SPUInit();
+    mutex.unlock();
+}
+
+void SPU::DeInit()
+{
+    mutex.lock();
+    // TODO - does it matter?
```
d3nwah, line 66: Uh, you tell us 😛

Source/AliveLibCommon/audio/Sequencer.cpp
```diff
@@ -0,0 +1,1552 @@
...
+}
+
+bool SeqPlay(s32 seqId, s32 repeats, bool stopDuplicateSeq)
+{
+    mutex.lock();
+    if (stopDuplicateSeq)
+    {
+        SPUSeqStop(seqId);
+    }
+    // SPUSeqSetVolume(seqId, 100, 100);
```
d3nwah, line 109: There's some commented out code here. If it's not needed, please remove :)

Source/AliveLibAE/Sound/Sound.cpp
sean-horton: This adds panning to the existing SDL audio rendering and is unrelated to the SPU emulation path. Since the existing SDL implementation is mono it's superior in the sense that it adds panning. I'm mentioning in case you don't want to modify the existing audio renderer otherwise I can remove the old panning.

tarceri:
You probably should be integrating Duckstation source code as the license is "CC Attribution-NonCommercial-NoDerivatives 4.0 International" note the "No Derivatives" this license is is incompatible with pretty much and free software license GPL, MIT, etc

rozniak:
> You probably should not be integrating Duckstation source code as the license is "CC Attribution-NonCommercial-NoDerivatives 4.0 International" note the "No Derivatives" this license is incompatible with pretty much any free software license GPL, MIT, etc

The license was changed on Sep 1st 2024 from the looks of it, and the last work on this was in June 2024 so would have been based on source that was licensed under GPL.

PR #1679, opened 2025-07-24
fix the playback feature

MrSapps:


Source/AliveLibAE/Fleech.cpp
```diff
@@ -1852,7 +1852,6 @@ void Fleech::ToIdle()
     mVelY = FP_FromInteger(0);
     mCurrentMotion = eFleechMotions::Motion_3_Idle;
     mNextMotion = eFleechMotions::eNone_m1;
-    sFleechRandomIdx_5BC20C++; // NOTE: unused variable removed - don't desync on RNG
```
mouzedrift, line 1855: this needs to stay or we need to remove it on master as well

Source/AliveLibAE/GameAutoPlayer.cpp
```diff
@@ -81,13 +81,6 @@ bool Player::ValidateObjectStates()
 
         BaseGameObject* pObj = gBaseGameObjects->ItemAt(i);
 
-        // hack fix - OG forgot to set the type of the throwable array in the ctor but it gets
-        // set in the VGetSaveState function so we assume that this is the case here to avoid desyncing.
-        if (pObj->Type() == ReliveTypes::eThrowableArray && reliveObjType == ReliveTypes::eNone)
-        {
-            continue;
-        }
-
```
mouzedrift, line 90: we don't set the type on master so this should stay or set the type on master

Source/AliveLibAO/Slig.cpp
```diff
@@ -1018,14 +1018,6 @@ void Slig::ToZShoot()
...
     field_114_timer = sGnFrame + field_174_tlv->mData.mZShootDelay;
     SetBrain(&Slig::Brain_ZSpottedEnemy);
-
-    // don't desync playbacks - this saved the result but never used it
-    // TODO: remove when we have a new AO recording without this math call
-    if (GetGameAutoPlayer().IsPlaying())
-    {
-        Math_RandomRange(3, 5);
-    }
-
```
mouzedrift, line 1028: this needs to be removed from master as well

Source/AliveLibAO/Slog.cpp
```diff
@@ -531,10 +531,6 @@ void Slog::ToIdle()
     mVelY = FP_FromInteger(0);
     mCurrentMotion = eSlogMotions::Motion_0_Idle;
     mNextMotion = eSlogMotions::None_m1;
-
-    // NOTE: unused field removed - keep the random call to prevent
-    // playbacks from desyncing on RNG
-    Slog_NextRandom();
```
mouzedrift, line 537: this needs to be removed from master as well

Source/relive_lib/ResourceManagerWrapper.cpp
```diff
@@ -427,14 +427,8 @@ void ResourceManagerWrapper::LoadingLoop(bool bShowLoadingIcon)
...
         if (bShowLoadingIcon && !bHideLoadingIcon && SYS_GetTicks() > startTime + k1Second)
         {
-            // master branch never hits the below code path so don't let the playback hit it
-            // if it does hit the loading icon Particle object will bump the object list count
-            // by 1 and desync the playback.
-            if (!GetGameAutoPlayer().IsPlaying())
-            {
-                // Render everything in the ordering table including the loading icon
-                ResourceManagerWrapper::ShowLoadingIcon();
-            }
```
mouzedrift, line 437: i think this should stay since this never gets hit on master but on beta it does

